PR #14, opened 2016-12-23
Provide support for installation of metadata tools

RayPlante:
This is needed integration into oar-docker where I'm adding a container to load the mongodb database with NERDm records converted from a PDL catalog.


RayPlante:
I note that this does not include schema changes that accommodate RMM's java data binding software; these will be separated into a different branch since such changes are expected to be temporary.

PR #15, opened 2017-02-22
File hierarchies, inventories, research topics, and field info

RayPlante:
(Please review: @deoyani)

This PR provides several updates to the creation and loading of NERDm records:

- File hierarchies are now supported in the NERDm schema; a sample dataset with a hierarchy ([janaf-hier.json](../tree/filehier/metadata/model/examples/janaf-hier.json)) can be ingested via oar-docker
- An inventory field is now defined by the NERDm schema, and `pdl2resources.py` will generate these (via the `-T` option). 
- `pdl2resources.py` adds a field called topic and fills it with improved values than what is in theme; these are based on the taxonomy defined in [theme-taxonomy.json](../tree/filehier/metadata/model/theme-taxonomy.json).
- Support is added to this repo for field info needed by the SDP.  Field info is machine-readable documentation about NERDm resource properties.  It is made available via the RMM.  This repo now includes field-info data file, [nerdm-fields-help.json](../tree/filehier/metadata/model/nerdm-fields-help.json) (with a JSON Schema-defined format).  [ingest-field-info.py](../tree/filehier/metadata/scripts/ingest-field-info.py) is a script to validating and loading the data into a MongoDB database.

Also, for [ingest-field-info.py](../tree/filehier/metadata/scripts/ingest-field-info.py), I added some general python code for validating and ingesting data into the MongoDB database; my thought that this could be reused by the RMM harvesters to update records.  A new script for loading NERDm records was created to use this same underlying code, [ingest-field-info.py](../tree/filehier/metadata/scripts/ingest-nerdm-res.py), so now the NERDm records are validated before being loaded into the database.

To test this PR, there are two things you can do:
- Check out the `filehier` branch and confirm that the unittests work.  The easiest way to do this is via the built in docker containers:
```
cd metadata
docker/run.sh testall
```
- Load the data into the MongoDB database via the [`filehier` branch of the oar-docker repo](https://github.com/usnistgov/oar-docker/tree/filehier) and see the results in the database.  To load, do this in your cloned copy of oar-docker:
```
git checkout filehier
cd rmm
docker-compose up -d
```
   (Be sure to fetch updates from github as well as delete any earlier created `rmm_*` docker images.)


RayPlante:
@deoyani Okay, taxonomy ingest support has been added (as well as into oar-docker).

RayPlante:
@deoyani A few more words about the `taxonomy` collection.  Each document in the collection represents a different term in the taxonomy.  Each document includes a property called `level`, indicating the level of the term in the taxonomy, where `1` is the top level (that would appear in the pull-down menu).  Thus to get that list, you can just do a query like so:
```
db.taxonomy.find({"level": 1})
```

Hope this works for us.

PR #17, opened 2017-05-31
feature: PDR's internal metadata server

RayPlante:
This PR adds the metadata server which is used to feed the PDR's internal, pre-publication landing page service with metadata from MIDAS.  This PR also locks in the spin-off of the [oar-metadata](https://github.com/usnistgov/oar-metadata) repository (which is now handled as a [git submodule](https://git-scm.com/book/en/v2/Git-Tools-Submodules)).

To build and execute the current unit tests, do the following:
```
git clone --recursive https://github.com/usnistgov/oar-pdr     # Note the use of --recursive
cd oar-pdr/docker
./run.sh testall
```


RayPlante:
requested wrong branch (master) to merge to.

PR #25, opened 2018-02-27
ODD-572: Provide unified, docker-based builds with version tracking

RayPlante:
This PR addresses issue [ODD-572 ("localdeploy Pt.2: provide unified docker-based builds in component repos")](https://mml.nist.gov:8080/browse/ODD-572).  It represents the second half of planned development of the localdeploy framework for building OAR applications.  All repos providing components to an `oar-docker` application, including this one, was updated with this framework.  It's key features are:
- Common mechanisms for building distributions from an OAR repository, via the common tools `makedist` (when prerequisites are natively installed) and `makedist.docker` (when they are not).  `localdeploy` uses the latter to build deployable distributions.
- Version tracking: a version string is generated based on `git` tags and branches as part of the (`makedist`) build process and allows that string to be injected into code.  `makedist` will also capture all of the dependency versions for the build and record them into a JSON file (of the form `*_deps.json`).
- All build output performed with `makedist.docker` will be owned by the user; this fixes a problem with the previous `localdeploy` where docker-based builds created files owned by root, making them a pain to delete.  
- Support for TravisCI automated testing

This PR primarily adds a `oar-build` directory containing generic framework files as well as new scripts to the `scripts` directory.  The latter files can be specialized for this package.  The original source for these tools can be found in the [`oar-build`](https://github.com/usnistgov/oar-build) repository.

The best way to test this is by building and locally running `oar-docker` applications; see the corresponding [`oar-docker` PR 95](https://github.com/usnistgov/oar-docker/pull/95) for instructions.


RayPlante:
This was tested by @haffonist  and @deoyani via [oar-docker PR95](https://github.com/usnistgov/oar-docker/pull/95); see their comments there.

PR #28, opened 2018-03-30
ODD-604: md service: don't lose external distributions

RayPlante:
This PR addresses [ODD-604 ("PDR-publish misses external distributions")](http://mml.nist.gov:8080/browse/ODD-604) which considers a problem handling a MIDAS SIP that includes distributions available from external URLs--i.e. not through the distribution service (including via s3.amazonaws.com and cloud front).  Such distributions would not get translated into the NERDm record via the metadata service.  Thus, these distributions would not be listed on the landing page.  

This scenario came up with a MIDAS SIP that was initiated prior to the MIDAS-PDR integration but is ready to publish afterward.  However, this problem would likely also arise when updating datasets that were published prior to the MIDAS-PDR integration.

Previously, NERDm record components were based exclusively on the datafiles present in the SIP input directories and not on the distribution metadata in the POD record (handled within the `nistoar.pdr.preserv.bagger.midas` module).  In this change, the POD distribution metadata is converted and retained; those that match data files in the input directories have their metadata enhanced (e.g. with size, checksum values).  The code for detecting when data files have been removed had to be updated.

### DIFF
--- a/python/nistoar/pdr/preserv/bagger/midas.py
+++ b/python/nistoar/pdr/preserv/bagger/midas.py
@@ -194,8 +194,12 @@ class MIDASMetadataBagger(SIPBagger):
                 log.info("Detected change in POD file (by checksum); updating.")
 
         if update:
+            # Note: setting savefilemd=True because there may be files that are
+            # part of the data set but don't exist in the input directories.
+            # These could be files available via external URLs or files preserved
+            # as part of a previous version.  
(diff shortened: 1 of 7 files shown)


RayPlante:
Tested on datapubtest against ediid=6376FC675D0E1D77E0531A5706812BC21886;  Distributions now appear.

PR #32, opened 2018-04-12
ODD-612: urlencode the filepath when converting to downloadURL

RayPlante:
This addresses [ODD-612 ("PDR-publish: files with URL-encodable parameters get listed twice")](http://mml.nist.gov:8080/browse/ODD-612) which describes how special characters in filenames were confusing the metadata service and causing them to be added twice, one with the characters URL-encoded and one unencoded.  The key to fixing this is ensuring that when converting between `downloadURL` and `filepath` properties, it is necessary to url-encode and decode.  

This PR has a companion [PR 25 to oar-metadata](https://github.com/usnistgov/oar-metadata/pull/25).  In this PR, the `bagit/builder` module was updated to encode the `filepath` when converting it to a `downloadURL` or a component ID (`@id`).  (The [oar-metadata PR 25](https://github.com/usnistgov/oar-metadata/pull/25) handles url-encoding.)

### DIFF
--- a/python/nistoar/pdr/preserv/bagit/builder.py
+++ b/python/nistoar/pdr/preserv/bagit/builder.py
@@ -7,6 +7,7 @@ import pynoid as noid
 from shutil import copy as filecopy, rmtree
 from copy import deepcopy
 from collections import Mapping, Sequence, OrderedDict
+from urllib import quote as urlencode
 
 from .. import (SIPDirectoryError, SIPDirectoryNotFound, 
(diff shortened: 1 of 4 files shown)


RayPlante:
This is tricky to test with oar-docker as it requires certain test data.  I will this all up on oardev.nist.gov for review.

RayPlante:
I've had the library and MIDAS teams review the changes on oardev.

PR #33, opened 2018-04-16
Revise Multibag spec to allow spaces in file names

RayPlante:
Bags produced by the preservation service were failing validation because the Multibag profile could not accommodate filenames containing spaces.  The profile specification was revised to handle this, and preservation service was updated to support the change.  The main changes in bags produced by the service are:
- The Multibag and NIST profile versions were changed to 0.3
- The bag names now contain the `mbag0_3` Multibag profile label
- The multibag tag files have changed in name and format:  `group-members.txt` and `group-directory.txt` are now called `member-bags.tsv` and `file-lookup.tsv`, respectively.  The format of the files are now tab-separated values (TSV).


RayPlante:
This was tested on oardev: 16 real-world submissions were sent through the preservation service, including those that revealed the bug being fixed.  Three of the datasets failed due to problems with the underlying POD file (i.e. not due to service bug).

PR #37, opened 2018-06-12
preservation service bug fix: correct failing async response message

RayPlante:
A large data submission triggered asynchronous completion of the preservation request, triggering a failure in submitting the HTTP response message.  This was due to a bug in an incorrect retrieval on the message from the preservation status object.  This has been corrected.  

I note that despite this failure, the preservation request should have still continued to successful completion, but it did not.  With further investigation, it was found that the `uwsgi` launching program was missing a required `--enable-threads` argument.  Apart from the context of PDR unit testing, this is handle by [`oar-docker`](https://github.com/usnistgov/oar-docker/); thus there is [a companion PR for that repository (PR 100)](https://github.com/usnistgov/oar-docker/pull/100) as well.

### DIFF
--- a/docker/pdrtest/entrypoint.sh
+++ b/docker/pdrtest/entrypoint.sh
@@ -15,7 +15,7 @@ function launch_test_mdserv {
     workdir=$PWD/_ppmdserver-test-$$
     [ ! -e "$workdir" ] || rm -r $workdir
     mkdir -p $workdir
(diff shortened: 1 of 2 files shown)


RayPlante:
This is tricky to integration-test without special preparations; I'm forgoing review for expediency.  I've tested this under `oar-docker` locally; I'll test it again on `testdata.nist.gov`.

PR #39, opened 2018-06-19
Data cart Angular CLI

skolli853780:


haffonist:
@skolli853780, the steps you provide are just to start the app on client side. How do I validate the functionalities?

haffonist:
To access oardev.nist.gov ( I beleive you meant oardev.nist.gov since datacart is not deployed on data.nist.gov), I need to be in Nist VPN which I am not.  So ask Ray/Gretchen to validate this validate this PR.  
This also shows how weak and lacking our testing environment is. ng serve should normally load fake data cart items without any reliance on external services to do so.

skolli853780:
@haffonist Instructions to go to data.nist.gov was to get EDIID, you can also get the EDIID from Dev or Test systems... without EDIID, you cannot access landing page as PDR Angular CLI version is not integrated with SDP Portal yet. using EDIID, you can access landing page and add files to data cart and then test data cart functionality.

PR #50, opened 2018-10-23
Fix/odd 658 pdr urls rewrite

deoyani:
This pull request has changes mainly for the issue 
[ODD-658](http://mml.nist.gov:8080/browse/ODD-658)
Added code to rewrite the url after landing page loading completes in browser. 
It removes /pdr part from URL to match with only /od/id pattern.

There are some changes with respect to 
[ODD-666](http://mml.nist.gov:8080/browse/ODD-666) 
Added new polyfill to jump to anchor section of url
e.g. http://localhost/od/id/<ediid>#dataaccess 
will jump to data access section of landing page.
The issue of going from metadata view to main page view section is still not completely resolved. 
But these changes are helpful for general anchors.

### DIFF
--- a/angular/src/app/landing/landing.component.ts
+++ b/angular/src/app/landing/landing.component.ts
@@ -8,8 +8,8 @@ import 'rxjs/add/operator/map';
 import { Subscription } from 'rxjs/Subscription';
 import { environment } from '../../environments/environment';
 import { AppConfig } from '../shared/config-service/config.service';
-// import { ErrorComponent } from './error.component';
-// import { ResComponents, DataHierarchy } from "./datacomponents.component";
+import {  PLATFORM_ID, APP_ID, Inject } from '@angular/core';
+import { isPlatformBrowser } from '@angular/common';
 
 interface reference {
   refType? : string,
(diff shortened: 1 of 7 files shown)


angular/src/app/landing/landing.component.ts
```diff
@@ -304,10 +298,30 @@ updateMenu(){
...
+           }
+        }
+      }
+    });
+  }
+
+  ngAfterViewInit(){
+    this.gotoSelection();
+    if (this.record != null && isPlatformBrowser(this.platformId) )  {
+      window.history.replaceState( {} , 'pdr/od/id/', '/od/id/'+this.searchValue );
```
RayPlante: It looks like there might be some confusion about meaning of the 2nd argument to `replaceState()`.  According to the [Mozilla developer documentation](https://developer.mozilla.org/en-US/docs/Web/API/History_API), the second argument is defined as:
> title — Firefox currently ignores this parameter, although it may use it in the future. Passing the empty string here should be safe against future changes to the method. Alternatively, you could pass a short title for the state to which you're moving.

To future-proof and otherwise avoid propagating the confusion, I would recommend changing the 2nd argument to an empty string.

PR #53, opened 2018-10-25
Fixed the unit and filesize display.

deoyani:
Issue reported in [ODD-614](http://mml.nist.gov:8080/browse/ODD-614)
This pull request mainly addresses the issue of files units and size display. 
1. Fixed as per requirement specified as 
when the size is 1000 bytes or greater that the value be converted to floating point with a metric byte unit such that:
* the number contain at least 3 significant digits
* there is 1-3 digits to the left of the decimal point
* there is a minimum of 1 digit to the right of the decimal point
* there is not more than 2 digits to the right of the decimal point

2. It also fixes some issue with URL rewriting when arkid specified in the url.

### DIFF
--- a/angular/src/app/landing/description/description.component.ts
+++ b/angular/src/app/landing/description/description.component.ts
@@ -205,16 +205,27 @@ addtoCart(fileName:string,fileSize:number,fileFormat:string,
  }
 
  display: boolean = false;
- /**
-   * Function to display bytes in appropriate format.
-   **/ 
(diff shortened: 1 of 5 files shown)


deoyani:
Updated this branch with some code clean up and indentation.
Fixed arkid issue.

RayPlante:
Requested changes from review have been made; subsequent "rider" fixes have been tested and verified under oar-docker.  Merging!

PR #56, opened 2018-10-26
Preservation Service: add support for X-HTTP-Method-Override for MIDAS

RayPlante:
This PR addresses a special issue with the MIDAS interface (and addresses a request from Chris).  MIDAS, as a Java application on the backend, uses the default Java 8 URL classes to connect to a service; these default classes do _not_, as is, support the lesser-used (though standard) HTTP methods like PATCH.  The simplest work-around is to use the [X-HTTP-Method-Override header variable](https://stackoverflow.com/questions/25163131/httpurlconnection-invalid-http-method-patch); however, this variable must be explicitly supported on the server.  This PR adds that support which allows a POST support to be converted to a PATCH.


RayPlante:
Self review: I tested this hot fix under oar-docker (see [oar-docker PR#104](https://github.com/usnistgov/oar-pdr/pull/56)).

PR #58, opened 2018-11-07
Environment variables reading before application startup

deoyani:
The issue with these environment setting came up in our deployment docker structure and the way environment variables were read in the code. 
Since PDR/angular code was using Resolver class to resolve incoming routes/urls. The process of reading variables from environment files and rendering page were running in parallel. 
This pull request solves one of the issue listed below as well.
[ODD-687](http://mml.nist.gov:8080/browse/ODD-687)

Testing:
After building and running application 
check:
1. sdp url in the header part pointing to the one which is set at runtime in the docker env
2. If record shows arkid as an identifier instead of 'doi' make sure the url points to the current pdr page
3. Export JSON link should point to 'rmmapi' set at runtime as well

In the process some other parts of code needs to be changed.
1. Rewriting URL in the browser after complete rendering of the page.
2. From 'Go To..' menu links so that when links are clicked page moves to  appropriate section.

Code is cleaned up so make sure all other functionality in general works fine.


deoyani:
The code for reading configuration has been updated and tested with local deploy.
Changes include:
Checking domain and adding /pdr for environment.json url
Removed reading process.env when app runs in browser as its for server side rendering.
Added few comments and updated error to be displayed as readable format.

RayPlante:
@deoyani: it appears I merged this oar-pdr PR too soon.  The fix did not work for int-pdr under the publish area: there I see the same link issues mentioned above.  Can you test the publish side under oar-docker?

(Note that I augmented your oar-docker PR to copy the changes to `createEnv.sh` to `publish/int-pdr/createEnv.sh`.)

RayPlante:
@deoyani:  quick tip for testing the LPS under publish:  after bringing up the system, access this test record:   https://localhost/od/id/0531A570681DB5D3A1EE2F169DD3B8CE1491

RayPlante:
@deoyani  False alarm.  I just did a test with publish, and the landing page worked fine--sorry about that.  It was under `demo` that it was going wrong for me.  I'll have to look into why.

PR #60, opened 2018-11-30
Fix MacOS-specific bug preventing makedist from working

RayPlante:
The `makedist` and `setversion.sh` shell scripts both contained commands that are not supported on MacOS--`realpath` and `sed -r`--causing `makedist` to fail on that platform.  These have been replaced with Mac-compatible alternatives.  (See also [oar-sdp PR# 199](https://github.com/usnistgov/oar-sdp/pull/199)).  

To test, just run `scripts/makedist` on a Mac to make sure it builds successfully.  It is not necessary to run the PDR application.

### DIFF
--- a/oar-build/_run.sh
+++ b/oar-build/_run.sh
@@ -23,7 +23,7 @@ LOGPATH=$PWD/$LOGFILE
 . $OAR_BUILD_DIR/_logging.sh
 
 function set_interactive {
-    shcmds=:`echo $SHELL_COMMANDS | sed -re 's/ +/:/'`:
+    shcmds=:`echo $SHELL_COMMANDS | perl -pe 's/ +/:/'`:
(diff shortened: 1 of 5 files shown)


deoyani:
Following error is displayed after running scripts/makedist

scripts/makedist.angular: line 34: realpath: command not found

RayPlante:
Ah, thanks for that.  Try again?

deoyani:
There are following errors coming up now after running scripts/makedist
Is there anything missing from command?

+ makedist.angular pdr-lps
Angular code path: /Users/dsn1/DeoWork/2018/11-Nov/oar-pdr-ray/angular
sed: illegal option -- r

RayPlante:
Eek!  Okay, try again?  thanks!

PR #61, opened 2018-11-30
Address event-stream vulnerability

RayPlante:
This PR addresses a vulnerability on the npm package dependency, event-stream, detected by GitHub (see GHSA-mh6f-8j2x-4483).  The remediation calls for locking the version of this package in `package.json` to 3.3.4 (as later ones are vulnerable).  

This PR also (re-)removes the `package-lock.json` file.  

I checked that our latest builds ended up only ever installing v3.3.4, so this change should have no effect on the application.  To test, simply confirm that the angular application still builds (starting with `npm install`).


deoyani:
Checked: install works fine. 
Build and test worked.

PR #63, opened 2019-02-05
ODD-712:  close bag's log on failure

RayPlante:
[ODD-712](http://mml.nist.gov:8080/browse/ODD-712) briefly describes how when a fatal failure occurs during preservation, the bag's internal log file does not get closed.  This causes problems removing the failed bag when the bag is on an NFS-mounted system, as it is on the test and production platforms. 

This small fix corrects this bug.  This PR also includes a fix to the `scripts/testall.python` script that ensures that it can run standalone.

### DIFF
--- a/angular/package.json
+++ b/angular/package.json
@@ -143,7 +143,7 @@
     "traceur": "^0.0.111",
     "ts-node": "~6.1.1",
     "tslint": "~5.10.0",
-    "typescript": "~2.7.2",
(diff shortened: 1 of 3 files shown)


RayPlante:
This is hard to test definitively except on the test and production systems that feature NFS-mounted disks.  Tests under oar-docker's demo app indicate that this PR does not break anything.  Will merge.

PR #67, opened 2019-03-19
Fix/internal pdr backend

deoyani:
This branch fixes issue with the internal PDR which was not working as it was not getting proper value for backend api. 
After adding small fix to service, it is fixing the problem. Localdeploy works fine.

### DIFF
--- a/angular/angular.json
+++ b/angular/angular.json
@@ -58,10 +58,7 @@
               "vendorChunk": false,
               "buildOptimizer": true,
               "fileReplacements": [
-                {
-                  "replace": "src/environments/environment.ts",
-                  "with": "src/environments/environment.prod.ts"
-                }
+                
               ]
             }
           }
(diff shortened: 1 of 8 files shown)


RayPlante:
I just ran under `publish` instead of `demo`, and the 1491 sample submission comes up fine, suggesting the problem I saw was with `demo`.  Looking into it...

RayPlante:
I found the problem with running under demo: a bit of code in `config.service.ts` that is meant to detect when running stand-alone and adjusts where the config file comes from; it was getting triggered unintentionally `int-pdr` is running under a special port, as it does under demo.  

This branch runs properly under publish, so I will accept the PR.

deoyani:
When you run locally (without docker local deploy, running directly npm run start etc) we do not need /pdr in the path to get assets/environment.json
It was a little hack we did to make sure things work fine with nginx proxy, /od/id/<recordid>
And /pdr/about page etc. 
In future we can find some better way to fix the path and routing issue.  
This is currently a simple fix for docker, nginx proxy , angular app for keeping /od/id path clear to only resolve given  <recordid>.

PR #71, opened 2019-04-03
Improve performance retrieving pre-pub metadata via asynchronous file examination

RayPlante:
This PR address issue [ODD-713](http://mml.nist.gov:8080/browse/ODD-713) which noted that when MIDAS submissions are very large, the mdserver can take a long time to process the submission and return a NERDm record.  This delay happens due to the time required to recompute checksums and to extract metadata from the files.  This PR reduces this delay by postponing this examination:  an initial version of the metadata is returned quickly to the client; meanwhile the mdserver launches a separate thread for doing the examination asynchronously.  Eventually, the file metadata will be updated with the correct checksum and additional enriched metadata; when the user next requests the metadata, the updated, enriched version will be returned.


RayPlante:
Tested and debugged extensively using oar-docker/demo, including with a large mock SIP.  Integration testing with MIDAS will be necessary on testdata/datapubtest.

PR #72, opened 2019-04-10
ODD753-757: Popup word wrap and direct download

chuanlin2018:
1. Fixed file detail popup dialog word wrap problem
2. Changed direct download function from http request to direct download because server side does not allow CORS. The code was commented out but not removed so we can easily revert once server side is fixed.

Makedist and docker local deployment were successful.


RayPlante:
works under oar-docker for me:  
 * I see descriptions now being properly wrapped.  
 * single file downloads from data cart is now via direct download

PR #76, opened 2019-05-21
New framework for setting and using configuration data

RayPlante:
This PR is part of a larger effort to fix server-side rendering.  Currently, exceptions occurring on the server-side are causing rendering to stop soon after encoding the NERDm JSON data into the landing page (via Angular's TransferState).  This larger effort is aimed at eliminating those errors as well as at making it simpler to make robust improvements going forward.  

In this step, this PR establishes a new framework for handling and using configuration information.  It takes into account the different runtime contexts (local development vs testing vs. production operation) and is described in the README.md file in [angular/src/app/config](/usnistgov/oar-pdr/tree/feature/newconfig/angular/src/app/config).  In particular, during development and testing, one will want to set the configuration via a local file; in production, the data must come from the configuration service, be used by the server but also transmitted to the browser to configure the browser-side application.

The new framework is implemented via a new (sub-) module in [app/config](/usnistgov/oar-pdr/tree/feature/newconfig/angular/src/app/config) (replacing the `config-service` in [app/shared](/usnistgov/oar-pdr/tree/feature/newconfig/angular/src/app/shared)).  When running the application in development (either as a browser-only app or with server-side rendering), the configuration is set via [src/enivronments/envirtonment.ts](/usnistgov/oar-pdr/tree/feature/newconfig/angular/src/environments/environment.ts).  When running under oar-docker, the server-side application reads the configuration from a file whose location is set by an environment variable.  (This is file is written and the environment variable set by the docker container that hosts the landing page service.)  The service side then transmits the configuration along with the rendered page using the Angular TransferState mechanism.  

The new framework introduces a new schema for the landing page service.  It will be retrieved from the config server via the name `pdr-lps`.  It adds some additional configuration parameters and renames others.  Documentation for the schema can be found in-line in [app/config/config.ts](/usnistgov/oar-pdr/tree/feature/newconfig/angular/src/app/config/config.ts).

### DIFF
--- a/angular/src/app/landing/landing.component.html
+++ b/angular/src/app/landing/landing.component.html
@@ -101,7 +101,7 @@
               <span *ngIf="isDOI"><i> <a class="font14"
                     href="https://doi.org/{{record.doi.split(':')[1]}}">{{record.doi}}</a></i></span>
               <span *ngIf="!isDOI"><i> <a class="font14"
-                    href="{{ confValues.PDRAPI }}{{record['@id']}}">{{record["@id"]}}</a></i>
+                    href="{{ cfg.get('landingPageService','/od/id/') }}{{record['@id']}}">{{record["@id"]}}</a></i>
               </span>
               <span *ngIf="checkReferences()"><br>
                 <span>Described in article: </span>
--- a/angular/src/app/landing/landing.component.html
+++ b/angular/src/app/landing/landing.component.html
@@ -158,7 +158,8 @@
               <description-resources [record]="record" [files]="files" [distdownload]="distdownload"
                                      [isInternal]="isInternal" [filescount]="filescount" [inBrowser]="inBrowser"
                                      [metadata]="metadata"></description-resources>
-              <metadata-detail [record]="record" [serviceApi]="serviceApi" [metadata]="metadata"></metadata-detail>
+              <metadata-detail [record]="record" [serviceApi]="serviceApi" [inBrowser]="inBrowser"
+                               [metadata]="metadata"></metadata-detail>
               <!-- <data-cart></data-cart>   -->
             </div>
             <div *ngIf="record.length !== 0 && inBrowser">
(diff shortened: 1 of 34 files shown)


angular/src/app/landing/landing.component.html
```diff
@@ -144,20 +145,23 @@ <h2>{{record.title}}</h2>
...
               </span>
             </div>
           </div>
           <div class="ui-g">
             <div class="ui-g-12 ui-md-12 ui-lg-12 ui-sm-12">
               <description-resources [record]="record" [files]="files" [distdownload]="distdownload"
-                [isInternal]="isInternal" [filescount]="filescount" [metadata]="metadata"></description-resources>
+                                     [isInternal]="isInternal" [filescount]="filescount" [inBrowser]="inBrowser"
+                                     [metadata]="metadata"></description-resources>
               <metadata-detail [record]="record" [serviceApi]="serviceApi" [metadata]="metadata"></metadata-detail>
```
chuanlin2018: Should we add [inBrowser]="inBrowser" here as well?

chuanlin2018: I understand that. I meant line 157 - I saw the MetadataComponent had @Input() inBrowser but seems the param was missing in line 157...

chuanlin2018:
DId you mean added "editEnabled" field to the record in the database?

angular/src/app/landing/landing.component.html
RayPlante: Ah, I see--sorry about that!  (The display did not make it obvious which line you were talking about.)  

Yes, you are right!!  It should be there; I'll add it real quick.  Thanks!

chuanlin2018: I enjoy reading your code :) very helpful for understanding server side rendering.

RayPlante:
No, sorry--I meant as a configuration parameter (as defined in [app/config/config.ts](/usnistgov/oar-pdr/tree/feature/newconfig/angular/src/app/config/config.ts) and made accessible to the app via an `AppConfig` instance).

angular/src/app/landing/landing.component.html
RayPlante: Thanks for the careful review!

chuanlin2018:
Just noticed "editEnabled" in the code. I will think about it.
For the endpoint for the editing service, yes I think we need that.

RayPlante:
FYI, oardev was updated with latest version.

PR #77, opened 2019-06-12
ODD773-Adding Google Analytics to PDR

chuanlin2018:
http://mml.nist.gov:8080/browse/ODD-773

I made following changes:
1. Added Google Analytics code to track pageviews.
2. Fixed about page router link.
3. Installed ts-node for e2e test.


RayPlante:
this is deprecated by #81.

PR #79, opened 2019-06-20
ODD-778:  Support ARK-ID-based EDI IDs

RayPlante:
This PR addresses [ODD-778](http://mml.nist.gov:8080/browse/ODD-778) which calls for supporting ARK-ID-based EDI IDs during the publishing process.  MIDAS will begin setting the `identifier` property in new POD records to an identifier starting with `ark:/88484/mds2-` and the PDR will adopt this as its local NERDm identifier.  A new heuristic was added for locating SIP directories for EDI IDs of this form.  To support updates to datasets that have already been published, the old style identifier will still be used.  

The update will require integrated testing with MIDAS on the datapubtest platform.


RayPlante:
Extensively tested under oar-docker (demo); further integration testing with MIDAS will happen on testdata.

PR #82, opened 2019-07-11
Fix docker build: install python module, funcsigs

RayPlante:
Something appears to have change in the python module dependency chain (probably related to [oar-metadata PR#31](https://github.com/usnistgov/oar-metadata/pull/31)), and now the python module, funcsigs, is an explicit dependency for the [NIST mulibag module](https://github.com/usnistgov/multibag-py).  In this PR, this module is now explicitly pip-installed in the `docker/pdrtest/Dockerfile`.  Without this change, the python tests fail.


RayPlante:
This change only affects Travis testing and distribution building. The following is sufficient for testing:
* seeing "All checks have passed" above
* ensuring scripts/makedist.docker runs successfully.

I have confirmed both; thus, as this is needed for other PRs, I am approving my self-review and will merge.

PR #90, opened 2019-07-17
Fix ODD vulnerability

chuanlin2018:
I merged 6 PR together (when trying to merge js-yaml-3.13.1 and fstream-1.0.12, it said already up to date. Seems already covered by other 4 PRs).

Ran PDR locally no problem.


RayPlante:
Thanks, @chuanlin2018.  I have two questions:
- Is the PDR actually using jquery?
- Since we're not using yarn, is there a reason to have a yarn.lock file?

chuanlin2018:
Let me check...

chuanlin2018:
I saw jquery.js under ultima-ng-5.2.4/assets/pages.
I don't think we need yarn.lock anymore. Should we remove it?

RayPlante:
> I don't think we need yarn.lock anymore. Should we remove it?

We should.  Can you also try removing jquery as a dependency in the package.json file?  @deoyani suggests that if it is need it should cause a problem at build-time.

chuanlin2018:
Sure.

chuanlin2018:
Removed yarn.lock and jquery. Tested in local and didn't see any problem.

PR #92, opened 2019-07-24
Fix ID validation failure in pre-pub metadata server

RayPlante:
In oar-pdr [v1.2.2](https://github.com/usnistgov/oar-pdr/releases/tag/1.2.2) (deployed as part of  v1.20 of oar-docker), the landing page service was failing to serve up pre-publication previews of the landing page; this was because the metadata server was still attempting to validate the ARK id for its check character, even though the service was configured not to for the new ARK ID format.  This error was traced to a use of the `BagBuilder` class within `nistoar.pdr.preserv.bagger.prepupd` where a `BagBuilder` instance was being instantiated without any configuration.  The default value of the `validate_id` config parameter is currently set to `True`; so the check was being erroneously applied.  This PR fixes this by providing an explicit configuration with this parameter set to `False`.

### DIFF
--- a/python/nistoar/pdr/preserv/bagger/prepupd.py
+++ b/python/nistoar/pdr/preserv/bagger/prepupd.py
@@ -371,8 +371,8 @@ class UpdatePrepper(object):
         if not nerdm.get('@id'):
             raise StateException("Bag {0}: missing @id"
                                  .format(os.path.basename(mdbag)))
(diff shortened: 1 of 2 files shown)


RayPlante:
This tested as a hot fix on datapubtest; the PR fixed the problem.  Will merge pending Travis testing.

PR #93, opened 2019-07-24
Fix failure updating submission with ARK SIP ID

RayPlante:
In oar-pdr v1.2.2, the preservation service would fail to preserve an update when the SIP uses an ARK-based identifier.  This was because the service was trying to cache a copy of the published NERDm record in a cache directory with the name `ark:/xxxxx/mds2-yyyy.json`  This PR corrects this problem, allowing such updates to be suportted.


RayPlante:
will resubmit using usnistgov branch.

PR #100, opened 2019-10-31
Feature/odd 761 customization

chuanlin2018:
This check in contains following changes:
1. Removed all hardcoded urls in auth service and customization service. Added customization api to environment.ts that the app will read when running in localhost outside docker.
2. When login first return 401 or 0, the app will redirect to sts url. The app will log/display other errors. I used alert() popup to display error code '200' or '401' or '0' because all console message will be clean up once window reloads. This is just for debugging purpose and will be removed once login works.


RayPlante:
This branch has been deprecated by [mdata-layout-new](https://github.com/usnistgov/oar-pdr/tree/mdata-layout).  A new PR will be created.

PR #103, opened 2019-12-02
Fixed tree table pre-select problem

chuanlin2018:
http://mml.nist.gov:8080/browse/ODD-812

Fixed the problem that only the first level tree table in datacart was checked after user clicks on the download all button.

Almost all code was changed because I changed indent from 2 spaces to 4.

The actual changes are the following functions:
            this.buildSelectDataArray(this.dataFiles);
            this.checkNode(this.dataFiles, this.dataArray);

### DIFF
--- a/angular/package-lock.json
+++ b/angular/package-lock.json
@@ -4726,9 +4726,9 @@
       "dev": true
     },
     "handlebars": {
-      "version": "4.1.2",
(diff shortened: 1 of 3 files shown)


RayPlante:
Deprecated by #109 after I botched a rebase.

PR #106, opened 2019-12-27
Odd 835 code cleanup

chuanlin2018:
This branch did the following changes:
1. Removed deprecated functions
2. Fixed app component unit test
3. Fixed broken social media link

PR #105, opened 2019-12-18
Mdata layout new

chuanlin2018:
This check in is for VDM-829, VDM-813, VDM-814 and VDM-828
http://mml.nist.gov:8080/browse/ODD-829
http://mml.nist.gov:8080/browse/ODD-813
http://mml.nist.gov:8080/browse/ODD-814
http://mml.nist.gov:8080/browse/ODD-828

Made changes for the following requests:
1. Fix unit test
2. Required fields - only authors is optional. Other edit fields are required - add red * next to the control buttons. Also add "* - required field" somewhere on the top.
3. Disable edit functionality for contact field for now.
4. Make Info icon bigger
5. In keyword pop up window, change the info displayed underneath the text box to "Please enter keywords separated by comma (* required)".
6. In edit pop up windows, disable Save button if edit content is empty or contains just spaces.
7. When user click on Submit button: pop up message to confirm: message TBD
8. In preview mode: enable submit and discard buttons (ODD-813)
9. In edit pop up windows, se if we can solve paste with style issue
10. In edit mode, disable hyperlink "view" under version
11. Show disabled "download all files" and "add all to cart" buttons
12. Keep track of updated record on refresh (VDM-814)
13. Disable the editing of the contact (VDM-828)

PR #107, opened 2019-12-27
Odd 786 resources by authors

chuanlin2018:
http://mml.nist.gov:8080/browse/ODD-786

This is the fix of PDR side. We need to use ASCII encoding of "&" and "=" in the request URL so SDP side can parse the parameters correctly. 

There are multiple places that spaces were removed. That was done automatically when I refreshed the code. You can ignore them. The part I changed was at the end of the file.


RayPlante:
Note: this builds on PR #106 (which itself builds on #105); don't merge until those others have been reviewed and merged.

chuanlin2018:
This is the URL pdr sent to sdp when I test locally:
https://data.nist.gov/sdp//#/search?q=authors.familyName%3DGarboczi,%26logicalOp%3DOR%26contactPoint.fn%3DZachary%20Levine&key=&queryAdvSearch=yes

Appeared to be working for me...

PR #111, opened 2020-01-15
ODD-835 addendum: Fix messaging when taxonomy fails to load

RayPlante:
@chuanlin2018 This is my recommendation for fixing the messaging/error handling when the taxonomy data does not load.


chuanlin2018:
Looks good to me.

PR #113, opened 2020-01-21
Fix/odd845 about page link

chuanlin2018:
http://mml.nist.gov:8080/browse/ODD-845

This branch fixed the about page link.

### DIFF
--- a/angular/src/app/app-routing.module.ts
+++ b/angular/src/app/app-routing.module.ts
@@ -10,7 +10,7 @@ import { NotFoundComponent, InternalErrorComponent } from './errors/errors.modul
 import { DatacartComponent } from './datacart/datacart.component';
 
 const routes: Routes = [
-    { path: '', redirectTo: 'about', pathMatch: 'full' },
+    { path: '', redirectTo: '/about', pathMatch: 'full' },
 
     // app paths
     { path: 'about',         component: LandingAboutComponent },
--- a/angular/src/app/frame/headbar.component.html
+++ b/angular/src/app/frame/headbar.component.html
@@ -19,7 +19,7 @@
         &nbsp;&nbsp;
         <span *ngIf="status" class="badge" id="appStatus" style="background-color:darkcyan">{{ status }}</span>
         <div class="header-links" style="float: right; vertical-align: bottom;">
-          <a href="/about" target="_blank"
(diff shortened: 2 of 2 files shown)


RayPlante:
Note that this fix doesn't quite do it:  if you do a curl on a landing page (which is what a bot will see without running the Javascript), the link comes out as `href="/pdr/pdr/about"`.  This is because the nginx proxy filters the output from the PDR server, substituting `href="/` whith `href="/pdr/`; see the the `location /od/id/` section of the [nginx config file](https://github.com/usnistgov/oar-docker/blob/563a7d27e29c882d3919fce944c99a26bad32ff5/apps/nginx-reverse-proxy/nginx-default.conf#L154).  Nevertheless, this is also the case on the current LPS deployed to production at the moment.  I suspect the proper solution will require some broader thinking that takes into account the issue of the error page URL that we discussed earlier today.  Thus, apart from one issue I'll raise in my next comment, I'm inclined to approve this patch.

RayPlante:
@GRG2 
I thought I would point out a difference between what we looked at in our developer's meeting today on the production system, and the pending release candidate.  On production right now (ver. 1.2.7), the landing page top navigation looks like this:
pdr-about-nav1 2 7

In the pending 1.3.0 rc that this PR would go into, the navigation looks like this:
pdr-about-nav1 3 0

In both cases, the link goes to the about page.  In the newer version, house icon link has a tool tip, "Go to home page".  In the later case, calling the About page the home page doesn't seem quite right.  If that's what we want to label it as (for now), then I think the tool tip should specify this as the   "Repository home page" (as distinguished from the portal's home page).  

So what would we rather go with:
   * Change the link back to "About"
   * Keep the "house" icon but change the tool tip to be more specific
   * Keep it as is (with the house icon and its current ambiguous tool tip).

Keep in mind that we just decided to add another link, "Contact".

chuanlin2018:
"About" seems more clear to me. I think I have never fully understood how this about page link works :) What I did was made it the same as it was in integration which was current production version. Ray or Deoyani may explain to me one more time sometime.
For the "Contact" link, when I have questions about a product, I usually click on "Support" menu item in their website. In the support page usually I can find phone number, email, FAQ, etc. Some support page allow me to submit message without pop up email client.

PR #116, opened 2020-01-24
508 compliance

chuanlin2018:
http://mml.nist.gov:8080/browse/ODD-792

Changes to color and button labels for 508 compliance.
Based on integration13.

### DIFF
--- a/angular/src/app/app-routing.module.ts
+++ b/angular/src/app/app-routing.module.ts
@@ -10,7 +10,7 @@ import { NotFoundComponent, InternalErrorComponent } from './errors/errors.modul
 import { DatacartComponent } from './datacart/datacart.component';
 
 const routes: Routes = [
-    { path: '', redirectTo: 'about', pathMatch: 'full' },
+    { path: '', redirectTo: '/about', pathMatch: 'full' },
 
     // app paths
     { path: 'about',         component: LandingAboutComponent },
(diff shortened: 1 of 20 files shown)


RayPlante:
@chuanlin2018 please resolve this conflict.

chuanlin2018:
Resolved. I saw you changed "integration13" to "integration". So "integration" is now restructured code.

RayPlante:
Did you really mean to choose pink here?
Spectacle z30096

RayPlante:
To answer my own question, I guess the answer is, yes, from looking at the file.  I applaud you for your boldness, but I think we need a different color here.  What is the recommendation this color change is addressing?

chuanlin2018:
In light background, a visited link is usually in dark pink color. So for dark background, I need to change it to lighter color for 508 compliance. I don't like this color either. Let's discuss in the standup meeting to pick a different color.

RayPlante:
Do we need to set a visited color?  Can we just keep it white?

chuanlin2018:
Sure. That's a good option.

RayPlante:
Cool.  Let's go with that.  Thanks!

chuanlin2018:
I like your idea. Actually a menu item should always be the same color. Will check in the change soon.

RayPlante:
FYI, I tested both under apps and publish as well as with editing enabled.

chuanlin2018:
Great. I tested in local oar-docker and everything appeared to be working fine.

PR #119, opened 2020-02-03
Updated social media link

chuanlin2018:
http://mml.nist.gov:8080/browse/ODD-853

Updated following links in the footer based on Peter's suggestion. Tested in local npm (not docker) since this is a fairly simple change:

Instagram

Looks like Travis failed while loading config data. Since this has nothing to do with this change, may be re-run Travis will fix itself?

### DIFF
--- a/angular/src/app/frame/footbar.component.html
+++ b/angular/src/app/frame/footbar.component.html
@@ -23,7 +23,7 @@
                   is external)</span>
               </span>
             </a></li>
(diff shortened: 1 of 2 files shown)


RayPlante:
@chuanlin2018 Can you fix this failed unit test--it's just testing a social media URL you just changed, so this should be easy to do.  --Ray

chuanlin2018:
Doing it now.

chuanlin2018:
Just checked in. Local test passed.

PR #120, opened 2020-02-11
LPS Restructuring Con't:  New Tools Component

RayPlante:
This PR is part of an ongoing effort at restructuring the angular Landing Page Service (LPS) app in an effort to improve maintainability and ease further evolution of the presentation.  Its overall goal is to distribute the current `LandingComponent` into several smaller apps that are aggregated together.  This PR, in particular, breaks out the right-hand tools menu into is own component, leaving the `LandingComponent` to handle just the main body of the landing page content.  While implementation of the menu is largely unchanged, it is a helpful precursor to replacing the tool bar with an implementation that is primeng-free (as the primeng components don't work with the server-side rendering).  

The change provided by the PR is summarized in the following diagrams.  
toolmenu-extraction

In particular, the changes include:
 * a new tools module (`ToolsModule`) containing a new `ToolMenuComponent` that encapsulates the current right-hand menu (also used as the pop-up menu while in mobile mode).
 * re-org of citation component, including 
    * placing it in its own `CitationModule`
    * splitting the layout into `CitationDescriptionComponent`, which handles the pop-up content, and `CitationPopupComponent`, which presents the content into a pop-up window.
 * the tool menu and citation component are now part of the `LandingPageComponent` composition, and removed from `LandingComponent`.  The latter now just handles the body content of the landing page.  
 * added the `NERDMResource` class to `app/nerdm/nerdm.ts` which handles more metadata transformations, including generating citation text.

(As a preview, the next restructuring PR will split the `LandingComponent` into the different navigable sections.)

To review, please:
 1. review this PR description
 2. scan the code differences as needed for any obvious issues
 3. run standalone with editing turned on and off to ensure proper layout
 4. run under oar-docker/apps to ensure proper integrated behavior

### DIFF
--- a/angular/src/app/landing/citation/citation.component.spec.ts
+++ b/angular/src/app/landing/citation/citation.component.spec.ts
@@ -1,0 +1,82 @@
+import { async, ComponentFixture, TestBed } from '@angular/core/testing';
+
+import { CitationModule } from './citation.module';
+import { CitationDescriptionComponent, CitationPopupComponent } from './citation.component';
+
+describe('CitationDescriptionComponent', () => {
+    let component : CitationDescriptionComponent;
+    let fixture : ComponentFixture<CitationDescriptionComponent>;
+
+    let setupComponent = function(citetext : string) {
+        component = null;
+        TestBed.configureTestingModule({
+            imports: [ CitationModule ],
+            declarations: [  ],
+            providers: [
+            ]
+        }).compileComponents();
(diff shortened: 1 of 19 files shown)


RayPlante:
Sorry, I accidentally created this PR from a personal branch (RayPlante) rather than from usnistgov.  I've pushed this same branch to usnistgov for your convenience.

PR #124, opened 2020-03-20
python: fix tests that rely on data.nist.gov

RayPlante:
This PR fixes some unit tests that sometimes fail or hang because they rely on accessing the production distribution service on data.nist.gov.  Instead, these unit tests now use a simulated distribution service running locally.  

This PR leverages the [PR#34 from oar-metadata](https://github.com/usnistgov/oar-metadata/pull/34) that modifies the conversion of POD records to NERDm which previously was too specific to data.nist.gov.  Conversion will now recognize distribution service URLs from any server, including a locally running one.


RayPlante:
This fix is need for all development since without it Travis testing is susceptible to intermittent failures.

PR #127, opened 2020-04-09
angular: landing page: fix detection of PDR landing pages

RayPlante:
When the `landingPage` metadata property points to an external web site, a button should appear at the top of the landing page that links to the landing page; if it's a PDR landing page URL, the button should not appear.  The code that detects an external URL was not properly recognizing PDR URLs using the new identifier pattern and the corresponding convention for the PDR landing page URL.  This PR updates to the implementation to recognize these URLs.

I also include a minor change that renders button as a traditional link when rendering on the server.

### DIFF
--- a/angular/src/app/landing/landing.component.html
+++ b/angular/src/app/landing/landing.component.html
@@ -73,7 +73,7 @@
                                 Visit Home Page
                             </button>
                         </span>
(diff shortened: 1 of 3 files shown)


chuanlin2018:
Reviewing...

chuanlin2018:
The code looks good to me. Tested locally, all looks good.

PR #128, opened 2020-04-10
Restore "View Metadata" function

RayPlante:
In the last PDR-LPS code restructuring iteration, the "View Metadata" item in the right-side tools menu stopped working.  This PR restores its functionality.

Note that metadata visualization is expected to change in further iterations of the code restructuring.

### DIFF
--- a/angular/src/app/landing/landing.component.html
+++ b/angular/src/app/landing/landing.component.html
@@ -80,16 +80,16 @@
 
             <div class="ui-g">
                 <div class="ui-g-12 ui-md-12 ui-lg-12 ui-sm-12">
-                    <div [hidden]="metadata">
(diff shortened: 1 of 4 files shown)


chuanlin2018:
I didn't see viewmetadata() in landing.component.ts be referenced anywhere. Not sure if we still need it. 
Another thing I noticed when I tested locally was that the fix to "Resources by Authors" was not in this branch. It was not in the integration branch either. So it's not caused by this fix. Probably it was missed out when we did restructuring.
Everything else looks good to me.

RayPlante:
Yes, I think `viewmetata()` is superfluous; I expect to remove it in the next restructuring iteration.

If you are good with this, can you approve the review?

PR #129, opened 2020-04-14
Added Appletouch icons

chuanlin2018:
http://mml.nist.gov:8080/browse/ODD-769

I reversed the link order in your readme file according to this article:
https://mathiasbynens.be/notes/touch-icons
so oldest iPhone will get the smallest icon.

Not sure how to test this locally. I don't have iPhone... There are free tools online but we have to deploy it first.


deoyani:
Works locally, once deployed can be tested on apple devices.

PR #133, opened 2020-04-30
publish bug fix: remove sha file entries from NERDm when files are remove

RayPlante:
There is a bug in the Python publishing code such that when a user removes a file and its corresponding `.sha256` file via MIDAS, MIDAS will remove both entries in the POD record; however, only the file would get removed from the NERDm metadata.  Consequently, the landing page would continue to show the sha files for files that had been removed.  This PR corrects this bug.


RayPlante:
I need to push this out as a hot fix, so doing self-review.  Confirmed fixed under oar-docker locally; will test on datapubtest.

PR #136, opened 2020-05-17
ODD-881: don't let file examination override download URL

RayPlante:
This PR addresses ticket [ODD-881](http://mml.nist.gov:8080/browse/ODD-881) that describes how during the publication process download URLs can get overridden when files are examined to extract extra metadata.  This was seen with new-style identifiers (e.g. `ark:/88434/mds2-7777`) in which a download URL that appears in the pod file like, say, `https://data.nist.gov/od/ds/mds2-7777/README.txt`, would get replaced in the NERDm file with `https://data.nist.gov/od/ds/ark/88434/mds2-7777/README.txt`.  While both URLs work in the PDR, the former is preferred.

### DIFF
--- a/python/nistoar/pdr/preserv/bagit/builder.py
+++ b/python/nistoar/pdr/preserv/bagit/builder.py
@@ -431,6 +431,10 @@ class BagBuilder(PreservationSystem):
 
     def _download_url(self, ediid, destpath):
         path = "/".join(destpath.split(os.sep))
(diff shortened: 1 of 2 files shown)


RayPlante:
This is a bit tricky to test.  I've tested this under oar-docker using the demo app and confirmed that download URLs are not getting overridden and preservation completes without issue.

PR #138, opened 2020-06-02
Customization: loosen validation restrictions on JSON submitted to /lp/editor

RayPlante:
The customization service (CS) was raising validation exceptions when the landing page service 
 (LPS) submitted author updates.  This was due in part because the latter was sending empty values for fields that the user did not fill in.  Leaving values empty, of course, is normal.  Rather than making the LPS remove unfilled properties, I thought it would be simpler to loosen the JSON schema definition the CS using to validate the changes.  The pubserver can scrub the changes more thoroughly to match the NERDm schema definition when it receives all the updates.  

I also tweaked the validation check to log all of the specific errors found in validation to aid in debugging such issues.


deoyani:
@RayPlante 

Thanks for updating this. Tested locally everything is working fine.

PR #141, opened 2020-06-10
Fix/odd866 bundle update

chuanlin2018:
Changes in this branch:

1. Changes to the download all confirmation popup
  a. Display the popup after bundle plan was received
  b. Display bundle details in the popup
  c. Display different message if bundle size exceed the threshold. The threshold is configurable.

2. Changes to bundle download UI layout
  a. Moved bundle download details (the green table) to the top and the warning message (the orange table) to be under the green table.
  b. Display progress bar for each bundle download. Removed spinner and the status display when downloading.
  c. Display bundle size
  d. Made error status clickable and display detail error message in a popup window
  e. Display overall download status
  f. Display estimated download time for each bundle
  g. Moved download status column to the left of Media Type column.
  h. Added a status icon to each file status
  i. Added background image to the title "Download Manager"
  h. Moved the three major control buttons into the title portion
 

More details at http://mml.nist.gov:8080/browse/ODD-866

### DIFF
--- a/angular/src/app/_helpers/fakeBackendInterceptor.ts
+++ b/angular/src/app/_helpers/fakeBackendInterceptor.ts
@@ -36,11 +36,11 @@ export class FakeBackendInterceptor implements HttpInterceptor {
       // }
 
       // Generate bundle plan
-      if (request.url.indexOf('_bundle_plan') > -1 && request.method === 'POST') 
-      {
-        console.log("Record saved...");
-        return of(new HttpResponse({ status: 200, body: bundlePlanRes }));
-      }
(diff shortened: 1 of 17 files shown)


RayPlante:
after I merged this and was testing a different branch, I noticed that we should change the pre-popup waiting message from "Zipping files..." (since that is not what it is doing).  In a separate branch, I changed it to "Preparing downloads...", and it is now deployed on testdata.  Let me know if you have any problems with it.

chuanlin2018:
Yes I just saw "Preparing downloads..." in testdata. Looks good to me. If you merge it to integration I will merge it to the new workflow. I am testing the merged version now.

PR #145, opened 2020-07-06
Feature/odd 855 customization api midas changes

deoyani:
This pull request addresses the changes in customization api as per new workflow design.


deoyani:
> This property name should be "user", not "userId", yes?
Following are the property names in customization api authentication and authorization sections but as per your request we can change to 'user' if that is used for getting permissions from Midas, it won't affect the customization ends.
{
userId: "TestId",
userName: "TestUser",
userLastName: "TestLast",
userEmail: "testuser@test.nist.gov"
}

RayPlante:
I'm talking about the JSON that gets sent to the mdserver endpoint.  Does this message conform to the proposal I emailed last week?

deoyani:
> I'm talking about the JSON that gets sent to the mdserver endpoint. Does this message conform to the proposal I emailed last week?

Short answer yes, your requirements are followed as per your email. In addition, the explanation says, Even though customization uses userId, I had set up mdserver request to use 'user' as it has been requested. In all other sections in customization userId is used as we had planned.

PR #122, opened 2020-03-05
Fix/odd857 new workflow

chuanlin2018:
http://mml.nist.gov:8080/browse/ODD-857

I made three checkins. The latest two were just minor fix and code clean up.

This fix made the following changes:

1. If edit is enabled, the app will be in edit mode when start up. It will not read nerdm record from transferstate but do authentication and load draft instead. It behaves the same as user clicks on the edit button in previous version.
2. The edit control buttons are: Preview/Edit, Done and Discard. 
    Preview/Edit button: In edit mode, Preview button will be active. All edit controls in editable fields will be active. In preview mode, Edit button will be active.  All edit controls in editable fields will be hidden.
    Done button: This button will be active if any of the editable fields was modified. When user click on Done button, the app send PATCH to backend to end the edit session. All buttons will be disabled.
    Discard button:  This button will be active if any of the editable fields was modified. When user click on this button, the app send PATCH to backend to reset the data to original state. The app will be set to edit mode.
3. If user is authenticated but loading draft return error, an "Ooops" message will display with detail info in email body.

This is how I test:
1. In app.module.ts, enable fakeBackendProvider to bypass the authentication
2. In environment.ts, set:
    useMetadataService: false,
    useCustomizationService: false
    editEnabled: true
3. Enable the following code in app/_helpers/fakeBackendInterceptor.ts:
            // Simulate loading draft
            if (request.url.indexOf('/customization/api/draft') > -1 && request.method === 'GET') {
                console.log("Interceptor returning sample record...");
                return of(new HttpResponse({ status: 200, body: sampleRecord }));
            }
4. Use http://localhost:4200/od/id/test1 for local testing
5. Use any other ediid to test error handling
6. In app/_helpers/fakeBackendInterceptor.ts enable "Simulate loading draft error" session to simulate loaddraft error. 

Just checked in with following changes:

messagebar.component.html : Added msg.text to the display.
messagebar.component.css: Changed warning color from black to dark orange.
Auth.service.ts: Remove re-direct parameter "editmode=true" because we don't need it for the new workflow (the app calls startEdit() at very beginning of ngOninit)
Customization.service.ts: Updated draftapi value to "pdr/lp/editor/" .
editstatus.service.ts: Updated one comment
Landingpage.component.ts: always call startEditing() if Editing is enabled.

PR #148, opened 2020-07-23
Fix/about pdr update

GRG2:
Completed update to documentation on PDR About page.    A couple styles were updated in consultation with @chuanlin2018.

Several links were added and or fixed which may require future maintenance.  Consider later changes to this.

Noted the link to API page included the # in url .   This branch was created from the #122  new-workflow


GRG2:
not sure what the Travis failure is due to

PR #149, opened 2020-07-27
ODD-764: Update metadata/preservation services to support customization workflow

RayPlante:
This PR addresses task [ODD-764 ("Update prepublication metadata server to accept updates")](http://mml.nist.gov:8080/browse/ODD-764) which address updating the Python components of the publication workflow to support user landing page customization.  In particular, this PR implements the ["Mark III" version of the MIDAS-PDR workflow](http://mml.nist.gov:8090/display/MID/MIDAS-PDR+Integration+Mark+III%3A+Landing+Page+Customization), in which MIDAS pushed updated POD records to the PDR (via a service api) as the user makes changes via MIDAS, and initiates landing page editing (via a separate api) so that the user can make changes there.  This PR introduces a new service, `pubserver`, to provide these new APIs.  During landing page editing, the `pubserver` service mediates between MIDAS and the customization service, converting metadata between POD and NERDm formats.  

Because of the change in the MIDAS-PDR workflow introduced in this PR, substantial changes were also necessary to the out layers of the preservation service.


RayPlante:
Tested extensively by the team and user-testers on datapubtest.  Will skip review.

PR #152, opened 2020-08-05
Fix/odd769 appletouch icon

chuanlin2018:
Peter's version of Appletouch icons checked in.


RayPlante:
@deoyani Can you review this so that we can include it in the 1.4.1 release?

PR #155, opened 2020-08-19
Fix dataset type labeling

RayPlante:
This PR addresses two problems related to the dataset type that appears above the title at the top of a landing page.  First there is a bug in the angular LPS code that prevents the proper type from being determined and thus displayed.  The result is that all landing pages were being marked "Data Resource", when the typical value should be "Public Data Resource" or "Data Publication" (the latter which should appear if the dataset has authors).  The second issue was that the '@type' NERDm property, from which this label is derived, was not getting updated when authors are added.  This latter issue is fixed in the python code used during the publishing process.  

There is a third minor, but unrelated. fix included.  (Yes, I've broken our normal rule about segregating fixes.)  I've added a little space in the rendering of the paper references that can appear near the top of the page.  

@chuan: I'm asking you to review, primarily so that you might look at the angular code changes for sanity check.  I've actually already deployed this to testdata so that I can see the fix with a variety of records.  If you wish to actually test, just load a landing page: if you see "Data Resource" instead of "Public Data Resource" (or some other value), then the fix is not working.


RayPlante:
PS: sorry for the plethora of "Merge" commits above; apparently I created my branch from main rather than integration.  Nevertheless, these "merge" commits have no file changes associated with them; thus, the changes shown when you click on "Files changed" above include only those intended as part of this PR.  These are captured in the last 5 commits.

chuanlin2018 (approved):
Yes I saw Public Data Resource in testdata.

PR #159, opened 2020-10-22
Fix/odd 908 913 914 search syntax

chuanlin2018:
This branch covers three Jira tickets.
ODD908: http://mml.nist.gov:8080/browse/ODD-908.
Changed "This data is referenced in" to "This data are referenced in".

ODD913: http://mml.nist.gov:8080/browse/ODD-913. 
"Resources by Authors" is now sending requests to SDP in following format: (example)
authors.fn="Z. H. Levine","A. R. Peskin","A. Holmgren","E. Garboczi" OR contactPoint.fn="Zachary Levine"

ODD914: http://mml.nist.gov:8080/browse/ODD-914.
"Similar Resources" is now sending requests to SDP in following format: (example)
keyword="ab initio","atomization enthalpy","basis functions","benchmark data",benchmarks,"bond angle".

### DIFF
--- a/angular/src/app/landing/data-files/data-files.component.html
+++ b/angular/src/app/landing/data-files/data-files.component.html
@@ -199,7 +199,7 @@
                 </span>
             </span>
             <span class="font14" *ngIf="isReferencedBy">
-                This data is referenced in :
+                This data are referenced in :
                 <span class="reflinks" *ngFor="let refs of record['references']">
                     <span *ngIf="refs['refType'] == 'IsReferencedBy'">
                         <br>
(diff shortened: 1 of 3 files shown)


RayPlante:
@deoyani Please look at the author search given in the description above.  Will this actually work as intended?

I believe the intention of the query is that we want any other record having an author or contact that matches one of the authors or contact in this record.  I'm concerned about the `author.fn` constraint.  Given that this value is stored in a fundamentally different way, will that search constraint work the way we want it to?

RayPlante (approved):
I'm approving this; although there is still room for improvement in the author search.  In particular:
* the author search searches for the full name exactly; thus, a resource with an author "Z. Levine" will not find a resource by "Zachary Levine".
* the search should only looks for resources where the authors overlap OR the contact is the same; it should also match those where one of the authors is a contact or vice versa.

PR #161, opened 2020-11-05
Feature/odd 936 add author scroll to bottom

chuanlin2018:
http://mml.nist.gov:8080/browse/ODD-936

This branch has the following changes:

1. The height of the popup window is fixed. 
2. The control buttons at the bottom were removed.
3. When user adds authors, it automatically scrolls to the bottom of the author list.


RayPlante:
Inadvertently closed.

deoyani (approved):
Worked locally as expected. The Add Author section stays on top all the time. It helps add new authors without scrolling up and down.

PR #162, opened 2020-11-17
Remove button height to use default

chuanlin2018:
http://mml.nist.gov:8080/browse/ODD-936

I removed button height (was set to 2em) from all "Add", "Save" and "Cancel" buttons in the edit popup dialog. Tested in Chrome and Firefox using local docker.

### DIFF
--- a/angular/src/app/landing/author/author-popup/author-popup.component.html
+++ b/angular/src/app/landing/author/author-popup/author-popup.component.html
@@ -14,17 +14,17 @@
         </div>
         <div class="ui-toolbar-group-right">
             <!-- Add author button -->
-            <button pButton type="submit" class="btn-add-author" style="background-color:rgb(221, 172, 9);"
+            <button pButton type="submit" id="btn-add-author" style="background-color:rgb(221, 172, 9);"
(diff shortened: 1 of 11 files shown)


angular/src/app/landing/author/author-popup/author-popup.component.html
```diff
@@ -14,17 +14,17 @@ <h2 class="popup-header">Authors</h2>
...
             </button>
 
             <!-- Cancel button -->
-            <button pButton type="submit" class="btn-cancel" (click)="activeModal.close('Close click')" label="Cancel"
+            <button pButton type="submit" id="btn-cancel" (click)="activeModal.close('Close click')" label="Cancel"
                 icon="faa faa-trash faa-1x icon-white" iconPos="left">
             </button>
 
             <!-- Save button -->
-            <button pButton type="submit" class="btn-Save" style="float:right; width:150px; background-color:green;"
```
RayPlante, line 27: Is this class/id name case-sensitive?  I think it is; if so, it doesn't match the stylesheet.

chuanlin2018:
Good catch! I also found the same issue in some other components. Will fix and check in.

chuanlin2018:
All fixed and tested in local docker. Please review again. Thanks.

PR #168, opened 2021-01-21
ODD-954: Require POD as input to preservation service

RayPlante:
This PR addresses [ODD-954 ("pubserver: require POD record as input to preservation request")](https://mml.nist.gov:8080/browse/ODD-954) which calls for a change to the preservation service API intended to prevent a common failure mode.  The fix now requires that the latest POD record to be provide as input to the preservation request to ensure that archived POD data is in sync with MIDAS, and most importantly, to ensure that the latest date held by MIDAS is actually valid.  The preservation service now processes the POD record just as if it was sent to the `/latest/` pubserver endpoint, and then proceeds with preservation; if the POD record is invalid, preservation is not carried out, and the client returned an error status.  

Test this PR via the [oar-docker PR#134](https://github.com/usnistgov/oar-docker/pull/134).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ services:
   - docker
 
 before_install:
+  - bash scripts/dhsetup.sh
   - cd docker && bash ./dockbuild.sh
 
 script:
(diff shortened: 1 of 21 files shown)


RayPlante:
Approved by @chuanlin2018 via [oar-docker PR#134](https://github.com/usnistgov/oar-docker/pull/134).

PR #170, opened 2021-02-11
Modified contact url link to be the same as in SDP

chuanlin2018:
http://mml.nist.gov:8080/browse/ODD-934

This fix modified the url of contact link to be the same as the one in SDP - from help/app-contact-us to help/contactus.

To test, run it in local docker and click on "Contact" at top right menu. Make sure the target is https://data.nist.gov/sdp/#/contactus.


RayPlante (approved):
Deployed to oardev; confirmed that help destination has been updated to match SDP.

PR #171, opened 2021-02-26
Fix/odd955-956-957 download all failure on mds2-2253, server side error and code clean up

chuanlin2018:
http://mml.nist.gov:8080/browse/ODD-955
http://mml.nist.gov:8080/browse/ODD-956
http://mml.nist.gov:8080/browse/ODD-957

This check in addresses above three tickets. 
For ODD-955 - Unable to Download-All on mds2-2253, it was caused by Google Analytics script because it takes too long to analyze the datacart page. Fixed this issue by moving GA call to after user confirm downloading. This is logically correct and solved timeing issue.

ODD-956 - server side error: fixed by adding inBrowser flag to non-server-side calls.

ODD-957 - remove dependency on EditStatusService in DataFilesComponent: the service was originally designed for other component but later on the design changed. Used a function call instead of doing a round-trip call. The un-used service was also removed.

Test it in local docker or deploy to oardev.

### DIFF
--- a/angular/src/app/landing/data-files/data-files.component.html
+++ b/angular/src/app/landing/data-files/data-files.component.html
@@ -1,5 +1,5 @@
 <!-- This bloack is for the data access session of landing page -->
-<div id="dataAccess">
+<div *ngIf="inBrowser" id="dataAccess">
     <div class="font14" [hidden]="metadata">
         <h3><b>Data Access</b></h3><br>
         
@@ -152,6 +152,8 @@
                                                 aria-hidden="true"></i>
                                             <span class="sr-only">Hidden text for 508 compliance</span>
                                         </a>
+                                        <p-progressSpinner *ngIf="rowData.downloadStatus == 'downloading'"
+                                        [style]="{width: '12px', height: '12px', 'margin-right': '.5em'}"></p-progressSpinner>
                                     </div>
                                     <ng-template #space_holder>
                                         <div style="padding-right: 0.4em;">&nbsp;&nbsp;</div>
--- a/angular/src/app/landing/data-files/data-files.component.html
+++ b/angular/src/app/landing/data-files/data-files.component.html
@@ -73,20 +73,25 @@
                                 data-toggle="tooltip" title="Download all files"
                                 [ngStyle]="{'color':getDownloadAllBtnColor(),'cursor':'pointer'}">
                                 <i class="faa faa-circle-thin faa-stack-2x" aria-hidden="true"></i>
-                                <i class="faa faa-download faa-stack-1x" aria-hidden="true"></i>
+                                <!-- <i class="faa faa-download faa-stack-1x" aria-hidden="true"></i> -->
+                                <i *ngIf="!addingToCart; else show_spinner1" class="faa faa-download faa-stack-1x"
+                                aria-hidden="true"></i>
+                                <ng-template #show_spinner1><i class="faa faa-spinner faa-spin faa-stack-1x"
+                                        style="color:#1E6BA1" aria-hidden="true"></i></ng-template>
                             </a>
                             <a id="routeToDatacart" target="_blank" [routerLink]="['/datacart', 'popup']"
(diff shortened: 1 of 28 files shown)


RayPlante (changes requested):
Tested on oardev after loading the mds2-2253 record:
ODD955: With download-all, I confirmed that it now successfully displays the cart and downloadable bundles (although it took a long time).  I was able to download all the bundles.  (We may want to up the file count limit.)  I take issue, though, with the new cart layout; see below. 
ODD956: Confirmed that exceptions no longer appear in server-side logs when displaying either the global cart or a download-all cart.
ODD957: Confirmed fixed based on code inspection, and the code still works (and status properly appears on the landing page).  

I see that the data cart layout has changed, seemingly unrelated to any of the issues this PR was meant to address, and I'm not really keen on the result.  First in the table showing the selected files, the columns are equal width; however their content are not.  Can we make at least the Download column and perhaps the Status column (depending on what goes in there) narrower so that Name column can be wider?  Also, what was the motivation for moving the Download column next to the name; I would prefer it be presented in a manner more consistent with the landing page and appear to the right of the size.

chuanlin2018:
Datacart layout is an easy fix. But I found another bug (also exist in current production): for the case of ark:/88434/mds2-2253, datacart won't be able to update the download status and the zip file names next to the file names because the process was unable to locate the files in the tree. Fixing...

chuanlin2018:
Finally, the following fixes are checked in:
1. The layour of the data tree.
2. Made the datacart work for param ark:/88434/ediid
3. Fixed performance issue - removed cart read/write in loops; combined functions to do less loops; etc.

chuanlin2018:
Not ready for deploying yet. Fixing another problem...

chuanlin2018:
You're welcome! I will revisit other functions sometime to make sure the documentation describes the functions clearly. That helps me as well.

As I said in today's meeting, the last check in the logic that counts the total number of downloaded bundle. It was off by one so the overall download status shows finished while there is still one downloading.

Tried to test it in local docker but some how local deployment took almost one hour. And SDP also took about 30 minutes. Finally I was able to test it in docker. Besides mds2-2253 not available in docker, everything looked good. The number of concurrent downloading bundle is now 3 instead of 2 (the new code counts from 0 to 2 I think).

chuanlin2018:
Hope this is the final fix. Tested in local docker. Thanks.

angular/src/app/landing/data-files/data-files.component.html
```diff
@@ -1,5 +1,5 @@
 <!-- This bloack is for the data access session of landing page -->
-<div id="dataAccess">
+<div *ngIf="inBrowser" id="dataAccess">
```
RayPlante, line 2: I think this use of `inBrowser` is unnecessary here as there is already one in place in line 52.  It unnecessarily removes content from view which will actually be noticeable by the user when the metadata includes a large file list.  

The use on line 52 guards against the use of the primeng tree component that does not work on the server side.  It also includes an "else" block which displays the "Loading file list..." which be seen for a short time in the browser when there is a large number of files.

RayPlante: I've removed this `inBrowser` (c92bd94) and tested on oardev.

RayPlante (approved):
I tested on oardev before and after my removal of the the `inBrowser` use mentioned above.  I confirmed that with it left in the display of the "Data Access" section is less optimal than with it taken out.  The server-side log messages now appear clean on both the landing page and the datacart.  I tested both the global data cart and "donwload-all"; downloading status appear to update successfully in both the status column of the cart page and in the "download status" box on the landing page.

angular/src/app/landing/data-files/data-files.component.html
chuanlin2018: Agree.

chuanlin2018:
Cheers! I tested it a little bit yesterday. Testing more...

PR #175, opened 2021-03-26
mdserver: fix renaming of files downloaded from publishing platform

RayPlante:
This PR addresses ODD-953 ("Datapub direct file download") which observes that files downloaded via the proto-landing page on the publishing platform (`datapub`) are being saved with names of the form, `download.`*.  A related symptom is that files with a comma in the name fail to download with the browser reporting a "Network error".  This is due to an incorrect format for the value being set for the `Content-Disposition` HTTP Header by the metadata server.  This corrects the value by setting it to `inline; filename="`_filename_`"`.

### DIFF
--- a/python/nistoar/pdr/publish/midas3/mdwsgi.py
+++ b/python/nistoar/pdr/publish/midas3/mdwsgi.py
@@ -362,7 +362,7 @@ class Handler(object):
         self.set_response(200, "Data file found")
         self.add_header('Content-Type', mtype)
         self.add_header('Content-Disposition',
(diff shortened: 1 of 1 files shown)


RayPlante:
As this is a super minor change that addresses a recent user issue on a pending publication, I'm going to treat this as a hot fix to datapub and skip review.  I've deployed and tested under oar-docker both locally and on datapubtest.  Downloading files with commas in the names is now possible, and downloaded files are now automatically being saved with their original name from the server.

PR #166, opened 2021-01-12
ODD-940:  Data Cart code restructuring

chuanlin2018:
This PR basically do the code restructure of data cart. The data cart page was split into three parts: cart controls, bundle plan and tree table. A new datacart class was created and most of cart related functions were moved from cart services into datacart class.

This PR changed the download all behavior - instead of popping up a new data cart window, now creating a new tab. 

This PR fixed the status sync between tabs - 
1. When files added to the cart in the landing page, they automatically showed up in the global data cart page if it's opened already (without refreshing the page). 
2. Any file downloaded in data cart page, the status in the landing page will be updated automatically  (without refreshing the page). 

This PR also added a new feature: display bundle download progress in the landing page. There will be one progress bar for each data cart (global data cart, download all data carts). Each progress bar can be manually closed once the download is completed or the tab is closed. The title of the progress bar is either "Global Data Cart" or the first 20 characters of the collection title.

To test, download the branch and run it in local docker as usual.

PR #185, opened 2021-05-25
Data Cart Restructuring Pt. 2

RayPlante:
As part of the on-going restructuring of the landing page code, PR #166 ("ODD-940: Data Cart code restructuring") was aimed at improving encapsulation of the code that manages the data cart.  That PR centered on the introduction of the `DataCart` class that controls the list of items currently in the cart.  This follow-on PR focuses on improving the encapsulation, particularly around the communication between components that use the cart as well as improving the separation of concerns for the different classes within the data cart module.  

This PR attempts to address the these main concerns:
 * several classes manipulated the state of the data cart by directly analyzing or manipulating a `TreeNode` data structure.
   * as `TreeNode` is part of the primeng `TreeTable` widget, that made multiple non-display classes dependent on primeng; this would make it harder to change to a different widget for the display.
   * for good encapsulation/separation-of-concerns in a MVC architecture, all such manipulations should go through the `DataCart` interface.
 * There are several communication channels and techniques used to communicate changes in the cart and the download status which could be consolidated around separation-of-conern principles.  (For example, the `BundleplanComponent` used the `DownloadService` to download data, but then used the `TreetableComponent` to tell the `DownloadService` that the download is complete, so that it can update the cart.)

This PR addresses these concerns primarily through these design changes:
  * The `DataCart` class is now watchable; thus all communications about changes in the cart, including the download state, goes through the `DataCart`.  
  * The role of the `CartService` has been redefined to be about obtaining an instance of a `DataCart`.  

I have also added a description of the Cart framework into [`datacart/README.md`](../blob/restruct/cartservice/angular/src/app/datacart/README.md) that attempts to describe the role of the key classes and explain the communication between them.


chuanlin2018:
I found three issues and two of them were fixed:
1. For ediid starts with "ark:", zip file name and download status do not show up in tree table (fixed);
2. Download status in the tree table is always green (fixed);
3. In the tree table, select/deselect the root of a dataset is not working properly. 
this.dataTree.children = [...this.dataTree.children] doesn't seem to be working. Function refreshTree() can be used as a work around but the tree table will fresh every time user select/deselect a treenode. So far I didn't find a better way to fix this.

RayPlante:
This is great--thanks for doing this.  The refreshTree() looks fairly inexpensive--nice!  Hopefully it won't cause any weird blinking.  I'll follow up with you on the next step.

chuanlin2018:
It does cause blinking :( So I didn't implement the change in this check in. You can try it to see if it's acceptable.

RayPlante:
@chuanlin2018:  This last commit (2f9e6d6) hopefully fixes the buggy selection behavior we were seeing (including # 3 in you list above).  Can you confirm?

chuanlin2018:
Will do.

RayPlante:
I'll note that this fix did not require the "who" capability that I shared in the [restruct/cartservice-selwho](https://github.com/usnistgov/oar-pdr/compare/restruct/cartservice...restruct/cartservice-selwho) branch, so I left that out.

chuanlin2018:
OK. I am reading your code while compiling locally.

chuanlin2018:
The problem seems to be fixed. I didn't see it happen again when running locally. Need a while to understand your code though:)

angular/src/app/datacart/treetable/treetable.component.ts
```diff
@@ -323,31 +331,50 @@ export class TreetableComponent implements OnInit, AfterViewInit {
      * @param event - change event
      */
     cartChanged(which) {
```
RayPlante, line 333: The functions that update the tree's selection status are:
* `propagateSelectionDown(node: TreeNode, selected: boolean)` -- this marks a node in the tree as selected (or unselected) along with all its child nodes.
* `propagateSelectionUp(node: TreeNode, selected: boolean)` -- this updates the selection status for a node's ancestor nodes (based on whether something below is being selected or unselected).

angular/src/app/datacart/treetable/treetable.component.ts
```diff
@@ -323,31 +331,50 @@ export class TreetableComponent implements OnInit, AfterViewInit {
...
         for (let item of this.dataCart.getFiles()) {
             node = this.dataTree.upsertNodeFor(item);
-            if (item.isSelected && (!this.tt || !this.tt.isSelected(node))) {
+            if (item.isSelected && (!this.tt || !this.tt.isSelected(node))) 
                 this.selectedData.push(node);
-                if (this.tt) this.tt.propagateSelectionUp(node, true);
+            if (this.tt) {
+                this.tt.propagateSelectionDown(node, item.isSelected);
+                this.tt.propagateSelectionUp(node, item.isSelected);
             }
```
RayPlante, line 345: This block looks for cart items that been added or changed.  If there is a new item in the cart, `upsertNodeFor(item)` will add a (leaf) node for it in the cart.  It's section status must be matched to what the cart says:  I call `propagateSelectionDown()` to (un)select the leaf node itself and then `propagateSelectionUp()` to update its ancestors accordingly.

angular/src/app/datacart/treetable/treetable.component.ts
```diff
@@ -323,31 +331,50 @@ export class TreetableComponent implements OnInit, AfterViewInit {
...
-        this.dataTree.cleanNodes(this.dataCart, (node) => {
-            if (this.tt && this.tt.isSelected(node)) {
-                this.tt.propagateSelectionDown(node, false);
-                this.tt.propagateSelectionUp(node, false);
-            }
-        });
+        // remove deleted items and adjust selection view accordingly
+        this.dataTree.cleanNodes(this.dataCart,
+                                 (node) => {
+                                     if (this.tt && this.tt.isSelected(node))
```
RayPlante, line 351: The `cleanNodes()` function removes nodes for items that were removed from the cart. This is the trickier part as I need to give it two call back functions that get called each time it removes a node.  The first (`removing`) get called before it is actually removed; this call-back unselects the node.  

The second (`refreshParent`) gets called after a node is removed and takes as it argument the node's parent node.  It's job is to refresh the ancestors selection status.  `propagateSelectionUp()` does this, but it really only works correctly if called on a leaf node; thus, I need to first find a leaf node below the parent (any leaf node will do) to pass to the function.

RayPlante (commented):
I don't know if this will help, but I'll try to explain what is going on (via inlined comments).  The `cartChanged()` function is where the view of selections is synchronized with what the cart says is selected.  The code is my least favorite because I had to use some functions of the primeng treetable component that are technically "public" but are nevertheless undocumented.  These functions, though, can update the selection status without triggering a selection event that would cause an event loop.  See below for the details.

chuanlin2018:
Just ran into couple of minor issues while going through your code: 
1. "Remove Downloaded" and "Remove Selected" did the same thing - both buttons call the same function.
2. Direct download in datacart was not working - problem with the parameters in this.dataCart.setDownloadStatus().

chuanlin2018:
I fixed the following:
1. "Remove Downloaded" - now calls right function.
2. Direct download in datacart
3. gaTrackEvent() in Google Analytics service - handle the use case that event = undefined while action = undefined. So far this never happened.

chuanlin2018 (approved):
Beautiful code. Everything looks great!

PR #187, opened 2021-06-30
Display no matrics available if nothing was downloaded

chuanlin2018:
No Jira ticket associate to this quick fix.

This fix is to display "Metrics not available" in the green menu in landing page if no file was downloaded regardless how many users were recorded in the metrics data.

Attached files show how the menu look like before and after the change.

01-before
02-after


RayPlante (approved):
Let's go for it!

PR #189, opened 2021-07-07
Get total file downloaded from record level metrics; check file level…

chuanlin2018:
Two changes in this check in:
1. Read "Total file downloads" from "success_get" in dataset level metrics. "Total dataset downloads" from "record_download";
2. In landing page, if no metrics data for current dataset, display "Metrics not available".


RayPlante (approved):
Appears to work!  Tested on oardev.

PR #192, opened 2021-07-09
Fixed missing metrics data;Display folder level downloads in tree table

chuanlin2018:
This check in fixed two issues:
1. Not all metrics data show up in the chart - somehow Object.assign() didn't append objects but replace object instead. It ended up we only have the last folder's data in the chart.
2. The total downloads of each folder didn't show up in the tree table - handleSum() was moved to a wrong location in one of previous check in.


RayPlante (approved):
Deploy this fix to oardev, and it works!

PR #193, opened 2021-07-09
Do not check metrics when edit enabled

chuanlin2018:
This check in added a guard to get metrics function to prevent loading metrics data when edit is enabled.

Since this is a fairly simple change I didn't test it in local oar docker...

### DIFF
--- a/angular/src/app/landing/landingpage.component.ts
+++ b/angular/src/app/landing/landingpage.component.ts
@@ -271,15 +271,17 @@ export class LandingPageComponent implements OnInit, AfterViewInit {
             }
         },
         (err) => {
-            let dateTime = new Date();
-            console.log("err", err);
+            console.error("Failed to retrieve file metrics: ", err);
         });                    
 
         this.metricsService.getRecordLevelMetrics(ediid).subscribe(async (event) => {
             if(event.type == HttpEventType.Response){
                 this.recordLevelMetrics = JSON.parse(await event.body.text());
             }
-        });
+        },
+        (err) => {
+            console.error("Failed to retrieve dataset metrics: ", err);
+        });  
     }
 
     get totalDownloadSize() {
--- a/angular/src/app/landing/landingpage.component.ts
+++ b/angular/src/app/landing/landingpage.component.ts
@@ -171,12 +171,15 @@ export class LandingPageComponent implements OnInit, AfterViewInit {
                 // id not found; reroute
                 console.error("No data found for ID=" + this.reqId);
                 metadataError = "not-found";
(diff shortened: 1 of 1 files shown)


RayPlante (changes requested):
While this guard solves the immediate problem while running on the publishing side, it does not address the general cause of the failure, which can be found in lines 278-281 of the modified code:  there is no error handler when retrieving record level metrics, so the exception that occurs rises to the top of the application, triggering the global "Oops" handler.  This could occur on the public side, too, given an ill-timed network or server error.  To fully guard against all possible retrieval errors, please add a an error handler to the record metrics request (as you do for the file metrics above it).  

Also line 275 (in the error handler for the file metrics request) shows the following:
```
     console.log("err", err);
```
I think you mean to use `console.error(...)`, yes?  I would recommend the following:
```
    console.error("Failed to retrieve file metrics: ", err);
```

chuanlin2018:
You are right. I originally thought about adding some error handling code but forgot later on... will do it today.

chuanlin2018:
Added dataset metrics error handling.

angular/src/app/landing/landingpage.component.ts
```diff
@@ -171,12 +171,15 @@ export class LandingPageComponent implements OnInit, AfterViewInit {
...
                 metadataError = "not-found";
-                //   this.router.navigateByUrl("/not-found/" + this.reqId, { skipLocationChange: true });
             }
             else{
-                // Get metrics 
-                if(this.inBrowser)
-                    this.getMetrics();
+                // Get metrics when edit is not enabled. Otherwise display "Metrics not available"
+                if(this.inBrowser){
+                    if(this.routerParamEditEnabled)
```
RayPlante, line 178: This should test `this.cfg.editEnabled` instead of `this.routerParamEditEnabled` since the latter only determines if the `editEnabled` query parameter was used.  The former properly indicates that we are running on the publishing side.

RayPlante (changes requested):
Sorry, I should have caught this before:  the guard code meant to prevent the call to `getMetrics()` needs tweaking.  When I ran the publishing app under oar-docker and retrieved a landing page without the "`?editEnabled=true`" parameter (i.e. I'm just previewing), I see the failure messages when attempting get the metrics.  That's good in that we know the changes to the error handling is working!  :-D  However, metrics retrieval is still being attempted. :-(

It should be a simple fix: see below

chuanlin2018:
Ah you are right. Will fix it.

chuanlin2018:
Tested locally with/without "?editEnabled=true" in URL.

RayPlante (approved):
Perfect!  Tested locally under oar-docker.  Thanks!

PR #194, opened 2021-07-14
Display dataset metrics instead of file metrics; Hide metrics display…

chuanlin2018:
This check in has two fixes:
1. In the top right green menu of the landing page, display dataset downloads instead of file download.
2. Hide dataset metrics block until metrics data is available (or error) so dataset metrics will not display "Metrics not available" then some metrics data.

Tested locally with fake backend. Testing in local docker...


RayPlante (approved):
Works great!  I was able to see both fixes when I deployed it to oar-dev.

PR #199, opened 2021-08-27
Refresh metrics data when user conduct a bundle download or direct do…

chuanlin2018:
http://mml.nist.gov:8080/browse/ODD-978

When user do a bundle download or direct download, the metrics data will be updated in certain period of time. The default time is 5 minutes. User can change it by alter the value of delayTimeForMetricsRefresh in environment.ts.

Two events trigger the getMetrics() function in LP: bundle download complete event and direct download click event.


chuanlin2018:
Thank you for the detail suggestions. I just checked in the following changes:
1. Made downloadstatus component independent from parent component's ediid. Instead it emits a set of item IDs of cartstatus whose download status is 'completed' (download percentage is 100).
2. For bundle download, landing page will check if the output of the downloadstatus component contains current ediid and refresh metrics only once per user download.
3. For direct download, the data-files component will emit a download status string so landing page will refresh the metrics data.

RayPlante:
Thanks for the updates.  I’m going to test later today (I hope).  

It’s not perfect, but it might be good enough.  The issue that’s still out there is that we have a flag, `dlStatus`, that indicates whether something from the landing page is being downloaded; however, we have N+1 (where N=number of files) ways of flipping that state simultaneously.  For example, what if the user initiates a bundle download and then clicks on several direct file download links?  If the completion signal flips the `dlStatus` before the bundle download completes, then will the metrics not get updated to reflect the bundle download?  

Here’s why it might be good enough:
- The above scenario may logically be rare (for users who understand what they are doing).
- There is a time delay in when the metrics is updated.
- There’s no real way to know when the direct download is really done (so it is signaled immediately after signaling the start); multiple direct downloads shouldn’t collide with each other.
- It’s not easy to signal once and only once when a bundle has finished downloading.

The ideal design would be one in which the `LandingPageComponent` receives an event when something has finished downloading.  The event identifies what was being downloaded, either a filename or cart identifier, and the component only receives that event for that file or cart only once (per user action).  I think for this to happen the `DataCartStatus` class, for one, would need to be a little smarter: when the data on disk changes, the `DataCartStatus` should detect what data has changed.

I don’t think it’s worth making it perfect right now, but I wanted to get these ideas down in case we need to revisit this.  Meanwhile, I’ll finish my testing.

chuanlin2018:
Make perfect sense. Will think about it.

RayPlante (approved):
Tested under oar-docker, and things appear to work based on the console messages; however, I didn't have any metrics data loaded, so I could not be sure.  I recommend confirming the behavior on datapubtest/oardev with metrics data loaded.

PR #201, opened 2021-09-15
Hide sha files

chuanlin2018:
http://mml.nist.gov:8080/browse/ODD-997

This change is to hide .sha files from the file list on the PDR landing page.

To test, run local docker and make sure no .sha file shows up in the file list.

Sample odid to test:
http://localhost:4200/od/id/66492819760D3FF6E05324570681BA721894
http://localhost:4200/od/id/3CF6F926D7309656E0531A5706810E0D1498

### DIFF
--- a/angular/src/app/landing/data-files/data-files.component.ts
+++ b/angular/src/app/landing/data-files/data-files.component.ts
@@ -251,14 +251,14 @@ export class DataFilesComponent implements OnInit, OnChanges {
 
         // Filter out sha files
         for (let comp of this.record['components']) {
-            if (comp.filepath && comp['@type'].filter(tp => tp.includes(':Hidden')).length == 0) {
-                if(!(comp['downloadURL'] && comp['downloadURL'].endsWith('sha256'))){
-                    node = insertComp(comp, root);
+            if (comp.filepath && comp['@type'].filter(tp => tp.includes(':Hidden')).length == 0 &&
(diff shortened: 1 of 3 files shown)


angular/src/app/landing/data-files/data-files.component.ts
```diff
@@ -237,24 +246,19 @@ export class DataFilesComponent implements OnInit, OnChanges {
...
-                if (cartitem) {
-                    node.data.isInCart = true;
-                    inCartCount++;
-                    node.data.downloadStatus = cartitem.downloadStatus;
-                }
-                */
-                if (node.data.comp['@type'].filter(tp => tp.endsWith("File")).length > 0) {
-                    count++;
-                    if (node.data.downloadStatus == DownloadStatus.DOWNLOADED) downloadedCount++;
+                if(!(comp['downloadURL'] && comp['downloadURL'].endsWith('sha256'))){
```
RayPlante, line 255: I would prefer the test be based on the file type (i.e. excluding ChecksumFile types) rather than on the file extension.  There can be checksum files that contain a different type (e.g. "MD5" or "SHA512"), and there can be checksum files provided explicitly by the author and should be considered the data file.  The publishing process determines which are which and sets the `@type` property accordingly.  

I can make and check-in this fix.

RayPlante (approved):
Tested on oar-docker before and after my tweak; both work.

PR #204, opened 2021-09-16
Restore display of "Loading file list..." message on server-side

RayPlante:
In the last round of restructuring (PR #196), a bug was introduced that prevented the momentary display of the message, "Loading file list...".   This message is generated on the server-side and is displayed in lieu of the file tree because this PrimeNG-based component cannot be generated on the server-side.  Determining whether to display this message is based on whether there are any elements in the `TreeNode[]` data structure that powers the file tree component. This structure can be generated server-side; however, due to a bug in the set-up logic, it was not getting generated.  Thus, on the server-side, it looked like this dataset had no files.  

This PR fixes this bug by ensuring that building the `TreeNode[]` structure--via `buildTree()`--happens on the server-side.  Because this function previously also updated the files' cart and download statuses, which requires access to the global data cart and, thus, can only happen browser-side, some changes in the logic for the latter was necessary.  Additional changes were needed to ensure that the dataset's overall cart and download statuses were updated consistently. 

I recommend testing this under oar-docker _or_ standalone in server-side mode (using `npm run build:ssr; npm run serve:ssr`).  Ensure the following for a dataset that includes files to display:
- When the page first loads, the message "Loading file list..." appears momentarily under the "Files" heading before being replaced with the file tree.  If not seen, confirm its non-appearance by examining the page returned by `curl`.  
- That visual indicators of file and folder statuses as well that for the overall dataset (i.e. the download and add-to-cart icons being either blue or green) update consistently.


chuanlin2018 (approved):
Confirmed in local docker:
1. Saw "Loading file list..." flashed before the file list showed up. Test case: JANAF
2. Add All to Cart buttons turned to green after all files added to cart
3. Download All buttons turned to green after all files were downloaded in global data cart

PR #205, opened 2021-09-21
Handle restricted access page

chuanlin2018:
http://mml.nist.gov:8080/browse/ODD-991

This change handles the restricted access page the way described in ODD-991. I added a feature that when user mouse over the "Download Data" button or access url link, the description will display.

To test, add a test record or modify an existing test record in environment.ts that contains at least one restricted public data component, say, test1. Then run it locally localhost:4200/od/id/test1.

### DIFF
--- a/angular/package-lock.json
+++ b/angular/package-lock.json
@@ -7975,9 +7975,9 @@
       "dev": true
     },
     "path-parse": {
-      "version": "1.0.6",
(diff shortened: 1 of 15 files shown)


RayPlante (approved):
Tested under oar-docker and reviewed by group; further review expected on oardev

PR #206, opened 2021-09-22
Hide time portion of the last modified date; Fixed Visit Home Page bu…

chuanlin2018:
http://mml.nist.gov:8080/browse/ODD-976

This checkin fixed the following problem:
1. Hide the time portion of the Last Modified field so it won't display "00:00:00";
2. Fixed the blue "Visit Home Page" button on the landing page - the click function was missing;
3. Fixed unit test in metadata component that disabled the unit test on all other components.

To test: the test case mds2-2154 does not exist in docker. I ran it locally with "useMetadataService" and "useCustomizationService" turned to true.


RayPlante (approved):
Works! (and interesting use of the pipe.)

PR #207, opened 2021-09-29
Remove hover function for restrict access page

chuanlin2018:
http://mml.nist.gov:8080/browse/ODD-1002

This change removes the hover action when user move the mouse over the "Download Data" button or the external link titles.


RayPlante (approved):
Works great!

PR #208, opened 2021-09-30
Changed the color of the JSON viewer

chuanlin2018:
http://mml.nist.gov:8080/browse/ODD-1001

This check in made the following changes to the JSON viewer:
1. Use light color for the title bar
2. Use blue color for the export button
3. Remove the background color of the expand level buttons ("1", "2", "3" and "View Full Tree") and hover background color to blue
4. Change text color in the title bar to black to meet 508 Compliance
5. Change text color inside the viewer to black to meet 508 Compliance.


RayPlante (approved):
Works and looks great!

PR #210, opened 2021-10-06
Fixed Safari error

chuanlin2018:
http://mml.nist.gov:8080/browse/ODD-1005

This error was because Safari has a strict requirement for the use of date pipe. It was unable to handle the date format in our lastModified field. The easiest fix is to just use the first portion of the field.

To test in Safari, in "Develop" menu choose "Disable Cross-Origin Restriction". Then change "lastModified" to different format and test "http://localhost:4200/od/id/test1":
1. If no lastModified field, "Last Modified" should not display
2. If lastModified only has date portion, LP should display as it is
3. If lastModified has both date and time, LP should only display the date portion.

### DIFF
--- a/angular/src/app/app.component.spec.ts
+++ b/angular/src/app/app.component.spec.ts
@@ -26,25 +26,38 @@ describe('AppComponent', () => {
         TestBed.configureTestingModule({
 
             declarations: [
-                AppComponent
+                AppComponent,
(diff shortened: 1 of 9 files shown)


chuanlin2018:
As we discussed, we only display the first 10 digit of the lastModified field.

PR #211, opened 2021-10-08
Apply date truncation to release date, too

RayPlante:
This is a follow-up to #206 and #210 in which we apply the same date truncation to the displayed "Release date" as we did to "Last Modified".


RayPlante:
@chuanlin2018 Do you have time to test this real quick.  Just run it standalone and view http://localhost:4200/od/id/test1; make sure both dates are truncated.

RayPlante:
That's fine.  I'll do it.  thanks a 10^6!

PR #212, opened 2021-10-15
Put metrics info in a separate component;Improved cartstatus to bette…

chuanlin2018:
…r handle metrics refresh

http://mml.nist.gov:8080/browse/ODD-993

Two major changes in this check in:
1. Removed metrics info from the green PrimNG menu and put it into a new component so each metrics info won't show up like a menu item. Might need further discussion on how metrics should show up in mobile mode. Or maybe we should get rid of the PrimNG menu so we have full control on how to display metrics.
2. Added a "Action" field to cartstatus structure so we have better handling of different action to the datacart. For now, we use "Set download complete" action to refresh metrics. 

Here is how the green menu and metrics look like:
Screen Shot 2021-10-15 at 4 21 40 PM


chuanlin2018:
Ready for review. Following are the changes:
1. Created a new component to display metrics info (hidden in mobile mode).
2. Added a "action" field to cartstatus class for better tracking of data cart activities. The main purpose of this change is to refresh the metrics data once user download one or more files.
3. In mobile mode, moved the menu button to the right side and kept it on top of the screen when user scroll.

RayPlante (changes requested):
Server side rendering generates this error:
```
LPS Application Error: ReferenceError: window is not defined
ReferenceError: window is not defined
    at LandingPageComponent../src/app/landing/landingpage.component.ts.LandingPageComponent.ngOnInit (/var/www/html/dist/server.js:146858:13)
    at checkAndUpdateDirectiveInline (/var/www/html/dist/server.js:23471:19)
    at checkAndUpdateNodeInline (/var/www/html/dist/server.js:24735:20)
    at checkAndUpdateNode (/var/www/html/dist/server.js:24697:16)
    at prodCheckAndUpdateNode (/var/www/html/dist/server.js:25238:5)
    at Object.updateDirectives (/var/www/html/dist/server.js:146659:438)
    at Object.updateDirectives (/var/www/html/dist/server.js:25026:72)
    at checkAndUpdateView (/var/www/html/dist/server.js:24679:14)
    at callViewAction (/var/www/html/dist/server.js:24920:21)
    at execEmbeddedViewsAction (/var/www/html/dist/server.js:24883:17)
Setting response status to 500
attempting reroute to /int-error
```
This is due to the use of the `window` object in `ngOnInit()` (which does not exist on the server).

angular/src/app/landing/landingpage.component.ts
```diff
@@ -148,25 +164,35 @@ export class LandingPageComponent implements OnInit, AfterViewInit {
         var showError: boolean = true;
         let metadataError = "";
         this.displaySpecialMessage = false;
+        this.CART_ACTIONS = CartActions.cartActions;
+
+        // Only listen to storage change if we are not in edit mode
+        if(!this.editEnabled){
+            this.cartChangeHandler = this.cartChanged.bind(this);
+            window.addEventListener("storage", this.cartChangeHandler);
```
RayPlante, line 172: This line needs protected with `inBrowser` guard!

chuanlin2018:
Ah, forgot about server side rendering... Will fix it today.

RayPlante:
I'm testing this now; meanwhile, consider my suggested tweak in previous comment.

chuanlin2018:
Finally I fixed all server side rendering errors.
In landingpage.component.ts, I disabled getMetrics() in server side because somehow server side does not recognize  event.body.text(). This could be just the problem when I test it in local. Will check in docker.

angular/src/app/landing/downloadstatus/downloadstatus.component.ts
```diff
@@ -45,19 +44,18 @@ export class DownloadstatusComponent implements OnInit {
...
     /**
      * Get keys of dataCartStatus. The UI uses it to display download progress
      */
-    get getKeys(){
-        return Object.keys(this.dataCartStatus.dataCartStatusItems)
+    get getKeys() { 
+        if(this.inBrowser) {
+            return Object.keys(this.dataCartStatus.dataCartStatusItems);
+        }else{
+            return null;
```
RayPlante, line 57: For some reason, my previous attempt at this comment did not show the line of code I was commenting on (`downloadstatus.component.ts`: 56).

I think it would be safer to return `[]` (an empty list) instead of `null`.

RayPlante (changes requested):

chuanlin2018:
Agreed. 
I am testing in docker to see if it's ok to run  this.getMetrics() in landingpage.component.ts in server side. 
Then I will check in the new code.

chuanlin2018:
Screen Shot 2021-11-24 at 3 28 03 PM
This part of code has not been changed so I suspect the error "event.body.text is not a function" is just a problem from my local...

chuanlin2018:
In the code I checked in I added inBrowser guard to this block to avoid the error from event.body.text(). That means in server side we cannot see metrics data...

RayPlante:
That's fine for now, I think.  We can revisit this later if it becomes important.

chuanlin2018:
I replaced "null" with "[]".

chuanlin2018:
Somehow my previous commit didn't show up...

RayPlante (approved):
Works now--thanks!  

One recommendation for the future (that is, let's not change it now):  the "get [function]" syntax in typescript (as in `get getKeys()` in `downloadstatus.component.ts`) means "make this function look and act like a class data field/property/variable"; thus, you should give it a name that is a noun (as in `get keys()`) rather than a verb phrase.  As a result, when you use (e.g. in `downloadstatus.component.html`), you would say, `let key of keys`.  This is more consistent with handling class variable.

chuanlin2018:
Make sense.

PR #213, opened 2021-10-19
LPS: Fix bug determining proper resource type label

RayPlante:
This PR fixes a bug that prevents the landing page from properly indicating the resource type above the dataset title.  Currently, all records are showing the type as "Data Resource", this value is displayed by default when the primary type is not recognized.  A resource's primary type is the first element of the `@type` property in the NERDm record.  The proper label is determined within the `ResourceIdentityComponent` via its `determineResourceLabel()` function.  This function was not accessing the `@type` property properly.  This PR fixes this bug and adds support for recognizing SRM datasets.  

This is fix is currently deployed to oardev.


chuanlin2018:
Do we have a record with "nrd:SRM" type for testing? I was unable to find it... All other types work fine.

PR #221, opened 2021-10-28
Disable useMetadataService for client side

chuanlin2018:
http://mml.nist.gov:8080/browse/ODD-1010

The error was caused by calling remote Web metadata service from client side.
This can be fixed by turning useMetadataService off in environment.prod.ts.

To test, run it in local docker.


RayPlante (approved):
works!

PR #223, opened 2021-11-02
Added nerdm document link

chuanlin2018:
http://mml.nist.gov:8080/browse/ODD-1014

Added NERDm document link to 'About this dataset' section. It links to https://data.nist.gov/dm/nerdm/.

### DIFF
--- a/angular/src/app/config/config.spec.ts
+++ b/angular/src/app/config/config.spec.ts
@@ -65,7 +65,7 @@ describe("config.AppConfig", function() {
         expect(cfg.locations.landingPageService).toBeDefined();
         expect(cfg.locations.landingPageService).toBe("https://data.nist.gov/od/id/");
         expect(cfg.locations.nerdmAbout).toBeDefined();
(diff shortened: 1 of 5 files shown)


RayPlante (changes requested):
The NERDm doc link needs to be provided in a platform-independent so that when on test, it points to testdata.nist.gov, etc.  The pdr-lps configuration includes a "nerdmAbout" parameter that provides the platform-specific link; so you should be able to have `AboutDatasetComponent` consult the `AppConfig` for this link.

RayPlante:
Also, BTW, the proper URL is `.../od/dm/aboutNerdm.html`; this is set as the default value in `AppConfig`.

chuanlin2018:
Tested in local.

RayPlante:
Looks like a small change is needed to `config.spec.ts`; see https://github.com/usnistgov/oar-pdr/runs/4120583005?check_suite_focus=true

chuanlin2018:
Ah, forgot to check in...

RayPlante (approved):
All good.

PR #224, opened 2021-11-16
Fix/odd1016 safari bug

chuanlin2018:
http://mml.nist.gov:8080/browse/ODD-1016

The problem was caused by Date.parse() function. Safari does not like "2017-01-01 00:00:00". Have to put a "T" between date and time: "2017-01-01T00:00:00".

The test1 sample in environment.ts has been modified for the testing of this case. 
To test, run app locally and browse http://localhost:4200/od/id/test1 using Chrome and Safari. Both should display the following line in version section:
    There is a more recent release of this resource available:    1.0.2

Screen Shot 2021-11-16 at 2 14 51 PM


RayPlante (approved):
Files look good.  We reviewed this as a group with a demonstration on Safari, and I confirmed that this still works under oar-docker and Chrome.  We'll next retest on Safari on test platform.  Approving to allow for release candidate.

PR #225, opened 2021-11-23
Handle file list display for restrcited public data

chuanlin2018:
http://mml.nist.gov:8080/browse/ODD-1017

This check in is based on the following logic:

If it's public data or not in publishing platform: normal display;
Otherwise if it is restricted data and not in preview mode: display special note and set the background color of the file list to grey;
Otherwise (restricted data and preview mode: hide the file list.

Also in current version, "editEnabled" has different meaning in different component. I made it consistent that variable "editEnabled" indicates publishing platform. And "editMode" indicates current edit mode.

I included restricted public data "test3" in environment.ts. "test1" is public data.
 
To test, change "ediid" in test1 to "test1" and set editEnabled to false in environment.ts and run the app locally.
Browse http://localhost:4200/od/id/test1, file list should display normally.

Browse http://localhost:4200/od/id/test3,  file list should display normally.

Set editEnabled to true then browse http://localhost:4200/od/id/test1 or http://localhost:4200/od/id/test3.  File list should display normally.

Set editEnabled to true then browse http://localhost:4200/od/id/test3?editEnabled=true, the note and file list should display on a grey background. Then click on "Preview" button, the note and the file list should be gone.

I believe these are all the logic that need be tested.


RayPlante (approved):
Works great!

PR #228, opened 2021-12-16
Replaced the file detail popup with row expand

chuanlin2018:
http://mml.nist.gov:8080/browse/ODD-1020

This check in replaced the file detail popup with the following:
1. When user mouse over a file name in the tree table, the row will expand and all file details will display there.
2. When mouse leave or user clicks on the file name, the file detail row will be closed if it's open.
3. An arrow next to the file name indicates the status of the file detail row.
4. A "copy to clipboard" button allows user to copy the hash value
5. A screenshot of the file detail row is attached here.
6. To test, run the app locally or in docker. Test case: http://localhost:4200/od/id/mds2-2437
Screen Shot 2021-12-16 at 5 20 23 PM
.

### DIFF
--- a/angular/src/app/landing/aboutdataset/aboutdataset.component.scss
+++ b/angular/src/app/landing/aboutdataset/aboutdataset.component.scss
@@ -83,7 +83,3 @@ $type-colors: (
   }
 }
 
-.highlight {
-    background-color: #4e6fff;
-    color: white;
-}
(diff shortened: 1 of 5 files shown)


chuanlin2018:
Actually I thought about exactly the same thing yesterday :) I was thinking about showing a pin next to the file name. When user click on it the whole mouse over function will be disabled.

RayPlante:
That would work better than my idea--it's more obvious.  Thanks!

RayPlante (approved):
Works great!

chuanlin2018:
Awesome!

PR #229, opened 2021-12-17
Support restricted public access in publishing workflow

RayPlante:
This PR addresses ODD-992 that updates the publishing system to support special processing of SIPs featuring restricted public access. SIPs from MIDAS are recognized as providing data under restricted public access when the POD includes files and an accessURL that points to the new Salesforce-connected Restricted Access Gateway; upon recognition, several bits of metadata are automatically attached to the record that are ultimately used by the Salesforce gateway.  It also applies two bits of specialized handling upon preservation and ingest:  the preservation bag files are sent to a special storage directory to be stored separately from the public data, and the file components are removed from the NERDm record sent to the ingest service.


RayPlante:
This was tested on datapubtest with MIDAS; skipping review.

Note that the corresponding MIDAS update is tabled until January.  We can still release our part; it will just go unused.

PR #233, opened 2022-01-24
Feature/odd 945 angular upgrade13

chuanlin2018:
http://mml.nist.gov:8080/browse/ODD-945

Finally we are at the latest version! To run this version:
node version: v12.20.0
npm version: 6.14.8
Angular Cli: 13.1.4

I have run client side, server side app and unit test locally without error. UI looks almost exactly the same as current version.


RayPlante (approved):
This was tested extensively by the group via oardev.  All good.

PR #240, opened 2022-03-29
Feature/odd 1025 science themes

deoyani:
@GRG2 
There are two json files in this repository under /proto section. 
/proto/forensics-updated-data/ForensicsAggregationCollection.json --> this record is a new collection record
/proto/forensics-updated-data/forensics.json ---> this contains all the updated records where the isProxyFor property is added and taxonomy schema dedicated to forensics use case.


RayPlante:
This has been replaced by #242 to address conflicts and integration with angular 13.

PR #248, opened 2022-05-21
LPS: always display DynamicResourceSets

RayPlante:
The NERDm schemas allow any type of resource record to include one or more `DynamicResourceSet` components.  Consequently, the dynamic search results should always be displayed--regardless of whether it is a `ScienceTheme` resource or not.  When the resource is a `ScienceTheme` resource, however, the display may be a little different.  Previously, the Landing Page Service code would only display dynamic search results for `ScienceTheme` records.  This PR fixes this.  

This PR also includes some changes to the `NERDResource` class.  Previously (on my advice), we moved a `selectAccessPages()` function from the `ResourceDataComponent` class to the `NERDResource` class; however, on closer inspection, that function included some manipulations specific to the `ResourceDataComponent`.  (More specifically, it annotated each selected component with some extra properties used for display.)  This PR splits this implementation between the two classes.  In addition, I've updated the `NERDResource` select functions to use its internal type matching functions to make it more robust.

### DIFF
--- a/angular/src/app/landing/resultlist/resultlist.component.ts
+++ b/angular/src/app/landing/resultlist/resultlist.component.ts
@@ -75,6 +75,7 @@ export class ResultlistComponent implements OnInit {
 
         let that = this;
         let urls = (new NERDResource(this.md)).dynamicSearchUrls();
+        console.log("ResultlistComponent: Found "+urls.length+" search url(s)");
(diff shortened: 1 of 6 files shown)


chuanlin2018 (approved):
Tested locally. Worked as expected.

PR #249, opened 2022-05-24
Fixed filters and added alt text to details toggle button

chuanlin2018:
This check in has the following fixes:

1. Fixed white space when close the Research Topics filter.
2. Added alt text to the dataset detail toggle button.

### DIFF
--- a/angular/src/app/landing/filters/filters.component.html
+++ b/angular/src/app/landing/filters/filters.component.html
@@ -24,7 +24,7 @@
                 <!-- Forensics Research topics -->
                 <div class="bottom-line" *ngIf="forensicsThemesTree.length > 0">
                     <div class="filter-checkbox">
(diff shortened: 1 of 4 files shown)


RayPlante (changes requested):
Alt text looks fine, but the filter expand/collapse behavior is still strange.  Here's what I'm seeing:

0. Initial State:
initial-state
1. Close Forensics Research Topics:
close-forensics
2. Close NIST Research Topics:
close-nist
3. Open NIST Research Topics:
reopen-nist
4. Reopening Forensics Research Topics returns to the initial state.  

And in a slightly different order, after the initial state shown above:
5. Close NIST Research Topics:
close-nist2

6. Close Forensics Research Topics:  looks just like step. 2 above

chuanlin2018:
Ah, I know what it is...

chuanlin2018:
The problem was Forensics research topics and NIST research topics shared the same variable.

PR #250, opened 2022-05-25
Got Author/Keyword filter dropdown to show up. Keyword dropdown still…

chuanlin2018:
… has text alignment issue in the dropdown.

Issue: when user clicked on Authors and Contributors or Keyword filter, the dropdown for authors only partially display and the dropdown for keywords does not show up at all.

Fix: set overflow of the filter panel to display instead of hidden. This makes the dropdown show up.

Issue to be fixed in the next release: the beginning and/or ending of the text of the dropdown of keywords will be truncated if keywords are very long. This seems to be a PrimeNg problem that might take a while to fix.

This check in also removed some console log that's for debugging purpose.

### DIFF
--- a/angular/src/app/landing/filters/filters.component.css
+++ b/angular/src/app/landing/filters/filters.component.css
@@ -126,3 +126,8 @@
 :host ::ng-deep .ui-tree .ui-tree-container {
     overflow: initial;
 }
+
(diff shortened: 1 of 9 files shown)


RayPlante (approved):
Works!

PR #251, opened 2022-05-27
Fixed keyword filter dropdown menu; Made Science Theme page 508 compl…

chuanlin2018:
…iant; Improved filters performance; other minor fixes

This check in did the following fixes:
1. Fixed keyword filter dropdown menu: set max length of each menu item for display.
2. Made Science Theme page 508 compliant;
3. Improved filters performance: made vertical scroll bars show up when user re-sizes the filter panel; Hide filter header bar text when filter panel becomes very narrow;
4. Fixed one logic problem: when user clicks on "Materials", filter panel shows 4, but result panel only display 2. - made the search case insensitive.
5. Removed the little gap between slider bar and search result panel.
6. Fixed Opps error in edit mode: it's waiting for other components to finish init. Added setTimeOut did the trick. I tested it locally.

Also tested server side rendering locally.

### DIFF
--- a/angular/src/app/frame/headbar.component.html
+++ b/angular/src/app/frame/headbar.component.html
@@ -7,8 +7,7 @@
       <div class="header-div">
         <span [ngStyle]="{'cursor': getCursor()}" (click)="openRootPage()" title="NIST Public Data Repository"
           class="header__logo-link" rel="home">
-          <img class="Fleft top_bar" src="./assets/images/nist_logo_reverse.png" alt="NIST Public Data Repository"
-            title="NIST Public Data Repository">
+          <img class="Fleft top_bar" src="./assets/images/nist_logo_reverse.png" alt="NIST Public Data Repository">
           <span class="Fleft"
             style="font-weight:bold;line-height: 16.5px;display: table-cell;vertical-align: text-top;color: #FFFFFF;font-size:14px;padding-left: 0.5em">PUBLIC
             <br> DATA REPOSITORY </span>
--- a/angular/src/app/landing/filters/filters.component.css
+++ b/angular/src/app/landing/filters/filters.component.css
@@ -11,6 +11,7 @@
     font-weight: 900;
     text-decoration:none; 
     padding-left: .5em;
+    cursor: pointer;
 }
 
 .spinner {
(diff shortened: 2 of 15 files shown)


RayPlante:
I'm concerned about the solution for item 6 above.  If the browser is under heavy load and slow to complete, will we just see this problem again?  It makes me wonder if `ngAfterViewInit()` is the right event hook method to use.  If it is (and it sounds like it should be), then perhaps this is the correct way to deal with it.  If you haven't already, you might double check the [component lifecycle docs](https://angular.io/guide/lifecycle-hooks).

At this point, I'm inclined to accept this fix if the testing works out.

RayPlante:
Okay, the more I think about the item 6 fix, the more it seems like it should be okay.

RayPlante (approved):
Tested on oardev; it looks like things work properly on the public side.  I will test the editing feature via datapubtest after merge.

chuanlin2018:
There is a variable _showContent that controls what time to display the LP content. This variable is controlled by editControl component. I believe when landingpage component tried to access the stickyMenu div, _showContent was not set to true yet. But it was so close that setTimeOut() function has enough delay time for editControl to finish.

I agree that there is a risk that the same error might occur under certain condition. But so far it seems to be fine. I will check the code again to see if there is a better way to handle this.


From: Ray Plante ***@***.***>
Date: Friday, May 27, 2022 at 4:31 PM
To: usnistgov/oar-pdr ***@***.***>
Cc: Lin, Chuan (Assoc) ***@***.***>, Author ***@***.***>
Subject: Re: [usnistgov/oar-pdr] Fixed keyword filter dropdown menu; Made Science Theme page 508 compl… (PR #251)

I'm concerned about the solution for item 6 above. If the browser is under heavy load and slow to complete, will we just see this problem again? It makes me wonder if ngAfterViewInit() is the right event hook method to use. If it is (and it sounds like it should be), then perhaps this is the correct way to deal with it. If you haven't already, you might double check the component lifecycle docs<https://gcc02.safelinks.protection.outlook.com/?url=https%3A%2F%2Fangular.io%2Fguide%2Flifecycle-hooks&data=05%7C01%7Cchuan.lin%40nist.gov%7C12a22d5e664a49b65c3608da401fe43e%7C2ab5d82fd8fa4797a93e054655c61dec%7C1%7C0%7C637892802959544976%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C3000%7C%7C%7C&sdata=UprW7AY%2FzFHgNOudA7zsMgvrjXtHEU8jHkZWvJ574wY%3D&reserved=0>.

At this point, I'm inclined to accept this fix if the testing works out.

RayPlante:
thanks!

PR #253, opened 2022-05-29
LPS: fix setMenuPosition() exception preventing editing

RayPlante:
The Landing Page Service fails to enter editing mode due to an exception in the `LandingPageComponent`'s `setMenuPosition()` function.  It was apparently getting called before the DOM was fully set for the child elements it needed access to; as a result, the function threw an exception when attempting to access a property of the undefined element.  This PR seems to fix the problem and do the right thing: adjusting the menu position as the page transitions between mobile and desktop modes.


RayPlante:
@chuanlin2018 This PR is currently running on datapubtest.  I'm just looking for a sanity check on this.

chuanlin2018 (approved):
Tested locally. Worked as expected.

chuanlin2018:
Also tested in datapubtest. Worked as expected.

PR #255, opened 2022-06-01
Fixed keyword filter issue.

chuanlin2018:
This check in fixed following issues:

1. When user selected more than one keyword in the keyword filter, nothing showed up in the search result. 
2. After above fixed, when user selected two keywords in the keyword filter then removed one of them, the search result will be empty.

#1 was because that the search string was not parsed. In SDP, it was parsed in the server side but here we need to parse it in the memory.

#2 was because the suggested keyword lookup table only contained the suggested keywords of latest selected keyword. Once it was removed, the app was unable to locate the original keywords for search.

To test, run the app locally and browse http://localhost:4200/od/id/mds9911.

### DIFF
--- a/angular/src/app/landing/filters/filters.component.ts
+++ b/angular/src/app/landing/filters/filters.component.ts
@@ -114,7 +114,7 @@ export class FiltersComponent implements OnInit {
     forensicsNodeExpanded: boolean = true;
     comheight: string = '50px'; // parent div height
     comwidth: string;  // parent div width
-    maxDropdownLabelLength: number = 30;
+    dropdownLabelLengthLimit: number = 30;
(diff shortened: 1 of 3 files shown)


deoyani:
@chuanlin2018 The sample data does not work at http://localhost:4200/od/id/mds9911
I tried running everything locally. Is the sample data checked in?

chuanlin2018:
Forgot to mention, when you test it locally, set useMetadataService: true in the environment.ts file. Then the app will use oardev as back end.

Screen Shot 2022-06-03 at 2 59 19 PM

deoyani (approved):
Tested locally and it works!

PR #256, opened 2022-06-02
Fix/metrics performance

chuanlin2018:
This version fixed following issues. It also included the change in PR#255 (keyword filter dropdown display nothing when user removed one selected keyword)

1. Fixed metrics info display: display spinner at right side of the green title bar before metrics data is ready.
2. Fixed the font in metrics info: defined the font family in style.scss to make it global. Set the font size locally.
3. Fixed "Visit Home Page" button color: I realized that it displayed science theme color all the time.
4. Made keyword suggestion list case-insensitive.
5. Display spinner in the search result section while loading: current version displayed "0 dataset found" first, then the number was updated.

### DIFF
--- a/angular/src/app/landing/data-files/data-files.component.html
+++ b/angular/src/app/landing/data-files/data-files.component.html
@@ -22,7 +22,7 @@
                         </span>
                     </div>
                     <ng-template #editDisabled>
-                        <div style="flex: 0 0 120px; text-align: left; padding-bottom: 0em;">
+                        <div style="flex: 0 0 130px; text-align: left; padding-bottom: 0em;">
                             <span><b id="filelist-heading">Files </b> </span>
                             <a class="faa-stack fa-lg icon-download" *ngIf="!editEnabled"
                                 (click)="downloadAllFiles()"
--- a/angular/src/app/landing/filters/filters.component.css
+++ b/angular/src/app/landing/filters/filters.component.css
@@ -14,12 +14,6 @@
     cursor: pointer;
 }
 
-.spinner {
-    position: relative;
-    width: 100%;
-    margin-top: 2em;
-}
-
 #clear-all {
     float: right;
     font-weight:900;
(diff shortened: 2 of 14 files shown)


deoyani:
@chuanlin2018 I tested all the changes except the metrics spinner etc, it is not obvious to me, can you confirm whether metrics API is called ? I do not see in the network calls in javascript console. Try using 8C401B5C708647B8E05324570681610A2081 as it has the data for metrics in oardev.

chuanlin2018:
Let me check.

chuanlin2018:
This is the url to get metrics data:
https://oardev.nist.gov/rmm/usagemetrics/files?exclude=_id&include=ediid,filepath,success_get,download_size&ediid=8C401B5C708647B8E05324570681610A2081

chuanlin2018:
I just ran it locally and I realized that line 206 in styles.scss need be commented out in order to see the spinner. I will check in the fix soon:
Screen Shot 2022-06-06 at 11 23 28 AM

deoyani:
> This is the url to get metrics data:
> https://oardev.nist.gov/rmm/usagemetrics/files?exclude=_id&include=ediid,filepath,success_get,download_size&ediid=8C401B5C708647B8E05324570681610A2081

Thats right but it does not seem to get called when I was testing this branch locally.

chuanlin2018:
The link was copied from console log. I added a console log right before the url was called...

chuanlin2018:
Just checked in the fix to the spinner css.

chuanlin2018:
It failed again in Python tests:
Screen Shot 2022-06-07 at 2 14 50 PM
I am not sure how to fix it except restart the process...

RayPlante:
It looks like we may have it an docker hub access limit; we may need to wait until tomorrow to rerun

RayPlante:
All tests have now passed.

deoyani:
Everything works! Metrics spinner needs testing on the server.

PR #257, opened 2022-06-13
Fixed duplicates in keyword filter lookup table

chuanlin2018:
This check in fixed some duplicates in the keyword filter dropdown list. For example in the attached screenshot - forensics and forensics(1) are actually the same.

Screen Shot 2022-06-13 at 4 26 14 PM


RayPlante (approved):
This appears to work with both #255 and #258.  Tested on oardev.

PR #258, opened 2022-06-14
LPS: Support access to previous versions

RayPlante:
The landing page offers links to older versions of a dataset record; however these link simply link to the latest version of the page.  These links were intentional placeholders until full support for accessing these old records.  This PR brings this support to the Landing Page Service.

Across all of the PDR components, this capability relies heavily on a new resolver service (provided via the new [oar-pdr-py](https://github.com/usnistgov/oar-pdr-py/) repo).  The purpose of this service is to resolve identifiers into different views of datasets in a variety of formats, be it NERDm, HTML, schema.org, etc.; this is accomplished via content-negotiation.  Thus, with this PR, the LPS uses the new resolver service to pull its metadata.  The API is slightly different from the RMM, so this PR adjusts for this as well.  

In addition, this PR effects the following changes:
 * The config data model was altered to (hopefully) make clearer the difference between API endpoints that should be used from typescript code (`APIs`) vs what appears in HTML links (`locations`), which should be used on the server (`serverSide`) vs. the browser.  See [config.ts](https://github.com/usnistgov/oar-pdr/blob/feature/version-access/angular/src/app/config/config.ts).
 * From the LPS angular app's perspective, all requests to the landing page are routed via the leading path, `/lps/`.
 * The URL for accessing a previous version has a form like this: `/od/id/ark:/88434/mds2-8888/pdr:v/1.0.0`.  Thus, the LPS has been updated to allow identifiers to include slashes.
 * The NERDm metadata for describing the version history has changed.  The presentation of the version history has been updated accordingly.

This PR is best tested via oar-docker using the [test/version-access branch](https://github.com/usnistgov/oar-docker/tree/test/version-access); this starts the resolver service and provides an updated coniguration.  To test, 
   * Build and run `apps` applicatoin from the [oar-docker test/version-access branch](https://github.com/usnistgov/oar-docker/tree/test/version-access).  
   * Make sure landing pages are created and rendered correctly in general for existing IDs
      * the `data/pdr/nerdm` directory contains a list of records that have been loaded into the RMM
      * make sure access to the datacart and the display of metrics still work
   * Make sure that a version is displayed in the top section
   * Open the version history display and click on a previous version:
      * make sure the displayed version matches the version you clicked on
      * ensure that the page includes a warning regarding a more recent version; make sure its link works correctly
      * open the version history to make sure that all versions (including later ones) are listed
      * click on the latest version to make sure the latest one resolves to the latest version.
   * Make sure the version display in the About section matches the one at the top of the page

The LPS should also be tested under `publish` application.


chuanlin2018 (approved):
Tested in oardev. Worked as expected.

PR #260, opened 2022-07-15
Fixed app error when there is no theme field in nerdm record

chuanlin2018:
http://mml.nist.gov:8080/browse/ODD-1051

This checkin fixes the problem that the landing page fails to load if the nerdm record does not have a "theme" field.

To test it locally, run the app with useMetadataService: true and useCustomizationService: true in environment.ts then browse the following url:

http://localhost:4200/od/id/75C6467F8B205670E05324570681995A1967

Or test it in local docker.


RayPlante (approved):
Works!

PR #262, opened 2022-07-25
Handle record that has no component - caused filter to fail

chuanlin2018:
Problem: when select "Record has" item in Science Theme collection, we got this error:

Screen Shot 2022-07-25 at 12 04 44 PM

Root cause: There is one record with no "components" field.

Solution: Fixed logic to skip the record that has no "components" field.

Test: Run it locally and browse http://localhost:4200/od/id/mds9911. Click on any of the three items under "Record has" in the filters section. Make sure the right side result list display the same number of files as indicated in the filter side.


RayPlante (approved):
Tested on oardev: filter now works.

PR #267, opened 2022-08-26
Fixed metrics url; Collapse filter when science theme landing page open

chuanlin2018:
Fixed the following two issues:
1. The link to metrics page broken: fix: treat /pdr/metrics/mds2-XXXX and /pdr/metrics/ark:/88434/mds2-XXXX the same.
2. Need to collapse the filter panel when science theme page first open.

mds2-2157 can be used to test metrics fix.

### DIFF
--- a/angular/src/app/shared/searchfields-list/searchfields-list.service.ts
+++ b/angular/src/app/shared/searchfields-list/searchfields-list.service.ts
@@ -20,7 +20,7 @@ export class SearchfieldsListService {
   constructor(private http: HttpClient,
     private appConfig: AppConfig,
     private cfg: AppConfig) {
-      this.RMMAPIURL = cfg.get("API.mdSearch", "/unconfigured");
+      this.RMMAPIURL = cfg.get("APIs.mdSearch", "/unconfigured");
     }
 
     ngOnInit(): void {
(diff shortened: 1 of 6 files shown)


angular/src/app/shared/searchfields-list/searchfields-list.service.ts
```diff
@@ -20,7 +20,7 @@ export class SearchfieldsListService {
   constructor(private http: HttpClient,
     private appConfig: AppConfig,
     private cfg: AppConfig) {
-      this.RMMAPIURL = cfg.get("locations.mdService", "/unconfigured");
+      this.RMMAPIURL = cfg.get("API.mdSearch", "/unconfigured");
```
RayPlante, line 23: This should be "APIs.mdSearch".  See [environment.ts](https://github.com/usnistgov/oar-pdr/blob/integration/angular/src/environments/environment.ts).

RayPlante: I'll probably fix this.

RayPlante (changes requested):
See inline comment regarding `mdSearch`; I'll probably fix this.

angular/src/app/shared/searchfields-list/searchfields-list.service.ts
chuanlin2018: Oh ya, I didn't test my last minute change... Go ahead fix it please.

PR #269, opened 2022-08-31
LPS: TopicComponent: guard against missing topic/theme properties

RayPlante:
This PR addresses ODD-1069 ("LPS: Topic component throws exception when scheme is missing") in which the implementation of `TopicComponent`'s `updateResearchTopics()` assumed the record has both `topic` and `theme` properties and that a `topic` always had a `scheme`.  This fixes the issue by adding guards that test for their existence.


RayPlante:
This issue turned out to be more critical than I thought:  on the public side, I could fix the offending records, but on the internal publishing side, users can save records that do not have topics or themes, yet.  Thus, I decided to go ahead and fix this one myself.

chuanlin2018 (approved):
Tested locally. Worked!

PR #282, opened 2022-09-29
Added media type lookup

chuanlin2018:
http://mml.nist.gov:8080/browse/ODD-1067

Display human readable media types in the file list.
Added following two types to the lookup json file:
"jpg": "image",
"gz": "compressed file archive",

This dataset has gz files in it: mds2-2190.
This dataset has jpg files: mds2-2645


RayPlante (approved):
Works!

PR #286, opened 2022-11-14
Add Google Analytics service to search result link

chuanlin2018:
http://mml.nist.gov:8080/browse/ODD-1076

Issue: in following page when user clicks on the "Home Page" button of "NIST Short Tandem Repeat DNA Internet DataBase", the "Oops!" error page will display:

https://testdata.nist.gov/od/id/pdr0-0001

Root cause: if landingPage field of a search result item is blank, and user click on "Visit Home Page", the Google Analytics service function will be called and failed because GA service was not defined in this component.

Solution: Add GA service to this component.


RayPlante (approved):
Works!

PR #301, opened 2023-01-10
Address json5 security issue

RayPlante:
This replaces #299, a dependabot PR that attempts to up-rev the json5 package from the insecure version 1.0.1.  This version of json5 comes in as a dependencies of the primary _dev_ dependency packages, regexp-replace-loader and ts-loader.  This PR addresses the json5 security issue by dealing with the versions of the packages that depend on the insecure version.  

The ts-loader dependency was uprev-ed to the latest version (9.4.2) which no longer depends on json5.  No effect was see on building the application.

Uninstalling and then re-installing regexp-replace-loader was enough to install 1.0.2 as its private dependency.  

As a bonus, the dev dependency on codelyzer was removed.  This package has been dormant for over a year and depends on angular v9.  We do not actively take advantage of this package at the moment.


RayPlante:
To get unittests passing consistently from GA, I had to include several python and docker fixes.

PR #304, opened 2023-01-13
use author fullname (fn) if it is available

bmaranville:
If an author record contains a fullname (fn) field, use that directly instead of reconstructing a full name from components (firstName, lastName and middleName).

It seems like fn should be preferred,  if it has been provided, given that the algorithms for constructing a full-name representation can depend on cultural, regional or even personal preferences.

### DIFF
--- a/angular/src/app/nerdm/nerdm.ts
+++ b/angular/src/app/nerdm/nerdm.ts
@@ -87,12 +87,18 @@ export class NERDResource {
         if (this.data['authors']) {
             for (let i = 0; i < this.data['authors'].length; i++) {
                 let author = this.data['authors'][i];
-                if (author.familyName !== null && author.familyName !== undefined)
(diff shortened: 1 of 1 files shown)


bmaranville:
To be honest all we really need for our use case is for the name string to fall back to author.fn if all the other names are undefined (in our user system only fn is defined) - and it might be less controversial than this change (which will affect a lot of existing citations!)

RayPlante:
I've had a mental sticky about this item for sometime, so we should go ahead and fix this.  I go back and forth on whether to prefer the fullname in the citation or just use it as a fallback.  The rational of the current implementation is to guarantee consistency in formatting.  Let's talk this through.  @GRG2 ?

chuanlin2018 (approved):
Tested locally. Worked as expected!

PR #306, opened 2023-01-19
Feature/odd 1083 rpa datacart

chuanlin2018:
http://mml.nist.gov:8080/browse/ODD-1083

This checkin modified the data cart to handle RPA request. Once the app detects a RPA request, instead of reading data cart from local storage, it requests data from the backend server, then creates a data cart structure and save it to local storage. The data cart works as usual.


RayPlante (approved):
Tested multiple times one testdata--all good!

PR #307, opened 2023-01-20
Fixed issues found in the test for release

chuanlin2018:
There is no Jira ticket for this checkin.

This check in fixed following issues:

Forensics Science Data page:

1. https://testdata.nist.gov/lps/pdr0-0001

When user clicked on "Clear All" in the filters panel, the search result was not reset and all filter categories were expanded (suppose only the category at the top expanded).

2. Extra dot at the end of the author list:

Barbara Guttman, Gregory Fiumara, Edward Sisco, Peter Vallone, Steve Lund, Melissa Taylor, Hans Soons..

Regular landing page:

1. The drag bar at the bottom of the data table missing sometime.

	Now it will show up when there are more than 25 items in the table - when scroll bar enabled, the drag bar will also show up.

2. In the case of large data set, when user expanded a folder, the table window will be set to minimum which only one line visible. The new behavior is that if the window is in minimum height when user expand a folder, it will be set to maximum height. Otherwise the height will be set to maximum.

3. The search glass icon does not have alt text.

### DIFF
--- a/angular/src/app/landing/data-files/data-files.component.ts
+++ b/angular/src/app/landing/data-files/data-files.component.ts
@@ -400,16 +400,23 @@ export class DataFilesComponent implements OnInit, OnChanges {
      * @returns 
      */
     treeTableToggled(event: any) {
-        if(this.files[0].expanded){
-            if(this.treeTableHeight == MaxTreeTableHeight) return;
+        //Set tree table's height based on the tree status
+        // If any node was expanded, set the height to max height
+        //Otherwise if files.length = 1, set height to min
+        let expanded: boolean = false;
+        this.files.forEach((file) => {
+            if(file.expanded) expanded = true;
+        })
+        this.isExpanded = expanded;
 
+        if(this.isExpanded){
             this.treeTableHeight = MaxTreeTableHeight;
-            this.isExpanded = true;
         }else{
-            if(this.treeTableHeight == MinTreeTableHeight) return;
-
-            this.treeTableHeight = MinTreeTableHeight;
-            this.isExpanded = false;
+            if(this.files.length <= 1) {
+                this.treeTableHeight = MinTreeTableHeight;
+            }else{
+                this.treeTableHeight = MaxTreeTableHeight
+            }
         }
     }
 
--- a/angular/src/app/landing/data-files/data-files.component.html
+++ b/angular/src/app/landing/data-files/data-files.component.html
@@ -69,7 +69,7 @@
                 [virtualScrollItemSize]="57" [globalFilterFields]="['name']">
                 <ng-template pTemplate="caption">
                     <div style="text-align: right; height: 20px;margin: 0px;padding-top: 0px;">        
-                        <i class="pi pi-search" style="margin:4px 4px 0 0"></i>
+                        <i class="pi pi-search" style="margin:4px 4px 0 0" aria-label="Search on file or folder name"></i>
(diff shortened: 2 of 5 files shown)


RayPlante (changes requested):
All fixes appear to work fine; in particular, these problems appear fixed:
  *  "Clear All" issue
  *  Extra period at end of facilitators list
  *  Sometimes missing file drag bar
  *  Search-glass icon alt-text

The behavior when folders are opened and closed is still peculiar.  As an example, I'm looking at /od/id/6998B81EF78777B2E05324570681D4DC1911.  After opening a few folders, I can expand the view of the file table to, say, 10 rows.  If next open or close a folder, the view portal will actually shrink.  This appears to be due to the view being larger than the value of `MaxTreeTableHeight` (perhaps as implemented in line 418?).  

I think the preferred behavior should that the view height should not automatically, unless it is less than the minimum height.  The max height should only be used to set the height _initially_ if the the number of top level files implies a height larger than the max value.

chuanlin2018:
I just changed the logic. 
If any node was expanded, and table height is minimum height, set it to maximum height.
Otherwise leave the table height as it is.

angular/src/app/landing/data-files/data-files.component.ts
```diff
@@ -401,21 +401,19 @@ export class DataFilesComponent implements OnInit, OnChanges {
...
+        //Otherwise leave the table height as it is.
         let expanded: boolean = false;
         this.files.forEach((file) => {
             if(file.expanded) expanded = true;
         })
         this.isExpanded = expanded;
 
-        if(this.isExpanded){
-            this.treeTableHeight = MaxTreeTableHeight;
+        if(this.files.length <= 1) {
```
RayPlante, line 412: It looks like this causes the view height to always shrink to the minimum if there is only one top-level item, even if that item is a folder.  

I think what you want is something like this (with no `else` section):
```typescript
if (this.filesCount <= 1  || this.treeTableHeight < MinTreeTableHeight) 
    this.treeTableHeight = MinTreeTableHeight;
```

RayPlante (changes requested):
This isn't quite there yet either.  I've deployed this branch to testdata, so have a look at https://testdata.nist.gov/od/id/mds2-2459.  (You'll have to do a Shift-Reload to get the updated code.)  Try slide the bar and opening folders in various combinations.

angular/src/app/landing/data-files/data-files.component.ts
chuanlin2018: In the case that there is only one top-level item, and that item is a folder, and the table height is MinTreeTableHeight, if user expand the top folder, do we want to expand the window accordingly? This is what the else section try to do.

chuanlin2018:
I modified the logic to handle following test cases:
Has only one file
54AE54FB37AC022DE0531A570681D4291851

Has two files:
26DEA39AD677678AE0531A570681F32C1449

Has multiple top folders
mds2-2437
5BD6911D381AB2E3E0531A57068151FA1869

Has only one top level folder
mds2-2459

Has many top level files
FF429BC1785F8B3EE0431A570681E858206

Has multiple top level files and folders
mds2-2116

RayPlante:
Yes--this is more like what I was thinking.  Testing now...

RayPlante (approved):
That's it!  Thanks for all the extra examples--very helpful.

PR #308, opened 2023-01-20
Fix LPS rendering failure when no components in metadata

RayPlante:
This fixes a bug discovered by @pjl1000:  the landing page will fail to render if the NERDm record does not include a components property.  (Note that this is not a required property.)  The fix requires the addition of a guard into  the `useMetadata()` function in `LandingPageComponent` that ensures the property's existence before attempting to calculate the total size of all files combined.


chuanlin2018 (approved):
PR #307 has the same fix for landingpage.component.ts but not the unit test.
Looks good to me.

PR #312, opened 2023-02-21
Fix/odd1086 minor fixes

chuanlin2018:
This check in solves the problems mentioned in Jira ticket: http://mml.nist.gov:8080/browse/ODD-1086.

Problem #1 seems only happens in edit mode while the two icons are disabled. To test this, set environment.ts to test mode then use this url: http://localhost:4200/lps/test3?editEnabled=true.

Problem #2 happens in both edit and public mode. Edit mode can be tested the same way as problem #1. Public mode can be tested using this url: http://localhost:4200/lps/ECBCC1C1301D2ED9E04306570681B10735.

Problem #3 also happens in both edit and public mode. 
  To test it in public mode, set environment to non-editing and use url: http://localhost:4200/lps/mds2-2116.
  To test it in edit mode, use http://localhost:4200/lps/test0?editEnabled=true. I added test0 to environment.ts for this purpose.

### DIFF
--- a/angular/src/app/landing/data-files/data-files.component.html
+++ b/angular/src/app/landing/data-files/data-files.component.html
@@ -1,7 +1,7 @@
 <!-- This bloack is for the data access session of landing page -->
 <!-- <div id="filelisting"> -->
 <div id="filelisting" *ngIf="displayMode != 'restrict_preview'"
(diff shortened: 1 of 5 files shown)


RayPlante (approved):
All issues appear to be addressed as seen under oar-docker locally.

PR #313, opened 2023-02-24
Bug fix: properly link to data via ghost record

RayPlante:
Currently, when someone updates a previously published record in MIDAS with new or updated files, the changes are sent to review under a new temporary record number--the so-called ghost record.  Meanwhile, the PDR can provide the draft landing page under that new number provided the submitted POD record contains a proper "replaces" property.  That landing page must link simultaneously to the new files and the previously published files that are not being updated.  Currently, there is a bug in the linking to the new files; the landing page service will only look for the files in directories associated with the new number when they actually are still stored under the new one.  This PR fixes this problem by adding the old record data directories to the list that the PDR checks for new/updated files.

Note that this bug was manifesting itself because of a separate bug in MIDAS: it was not always setting the "replaces" property.

### DIFF
--- a/python/nistoar/pdr/preserv/bagger/midas3.py
+++ b/python/nistoar/pdr/preserv/bagger/midas3.py
@@ -180,7 +180,7 @@ class MIDASSIP(object):
             if replid and (replid.startswith('doi:') or replid.startswith("http")):
                 replid = None
         if replid:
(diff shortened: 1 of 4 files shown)


RayPlante:
tested via a hot fix.

PR #316, opened 2023-05-26
Fix/rpa ssr

chuanlin2018:
This checkin fixes two problems:
1. The Opps error when RPA datacart starts - server side rendering problem.
2. The Opps error after user confirms a bundle download - caused by missing resTitle.

I disable the cartService.getRpaCart() call for server side rendering. Also use "No title..." if resTitle is not available.

Tested in local but not in docker. I need some time to install python 2.7 in my machine (currently running python3).


RayPlante:
This should have been made a PR against the `fix/ODD-1083-rpa-datacart`; see [here ](https://github.com/usnistgov/oar-pdr/compare/feature/ODD-1083-rpa-datacart...fix/rpa-ssr)for difference with #306.

chuanlin2018:
I see. I will undo this commit and make change to fix/ODD-1083-rpa-datacart instead. Thanks.

RayPlante:
All changes that are in this PR are also in #306; closing this one without merging.

PR #318, opened 2023-06-08
LPS: set rel="noopener" to links with target="_blank"

RayPlante:
This PR is in response to a security scan of  the datacart page serving RPA requests.  The scan recommends adding `rel="noopener noreferrer"` to any `<a href="...">` that also has `target="_blank"`.  This would apply to any link generated in the PDR angular application.  

There are two disadvantages to making this recommended change:
  *  when the link points to a downloadable file, an access page, or a home page, the use of `noreferrer` prevents target sites from learning that the PDR is driving users to that site, including when that site is the PDR itself.  Note that in most cases, when this type of link does not point to a PDR URL, it points to another `nist.gov` server; thus, "noreferrer" mainly hurts ourselves. 
  *  when the link points to a PDR angular URL, use of `noopener` prevents angular from leveraging the linkage to the referring app.  

Thus, in response, this PR applies the following changes to templates with links with `target="_blank"`:
  *  if the link points explicitly/definitively to a PDR URL, no change is made.  
  *  if the link points to a downloadable file, an access page, or a home page, `rel="noopener"` is added to the link tag.
  *  if the link points definitively to an external site (e.g. a social media site), `rel="noopener noreferrer"` is added to the link tag. 

I have tested this under oar-docker and confirmed that the new attribute appears and that the links still work.

### DIFF
--- a/angular/src/app/datacart/treetable/treetable.component.html
+++ b/angular/src/app/datacart/treetable/treetable.component.html
@@ -63,7 +63,7 @@
                     </td>
                     <td [ngStyle]="bodyStyle(sizeWidth)">
                         <div *ngIf="rowData.cartItem;else space_holder" style="display:inline;color: green;" >
-                            <a *ngIf="rowData.cartItem.downloadStatus != 'downloading'" href='{{rowData.cartItem.downloadURL}}' target='_blank' download="download" data-toggle="tooltip" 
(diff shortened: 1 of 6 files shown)


chuanlin2018 (approved):

chuanlin2018:
Sorry I forgot about this PR. Should I merge it?

PR #319, opened 2023-06-09
Fix/odd 1110 rpd

chuanlin2018:
…ict public data

http://mml.nist.gov:8080/browse/ODD-1110

To test locally, use http://localhost:4200/lps/test3. This restrict data record contains two data files. One has download url and the other has not.

The last column of the data table should be named" "Access". The row that does not have the download url should display a lock and link to the accessURL of the page.

The expand function of "About downloading restricted public data" was fixed.


RayPlante (approved):
Tested on testdata combined with #306 branch, confirmed all changes.

PR #321, opened 2023-07-11
Handle page view for SPA

chuanlin2018:
Google Analytics is not notified when we route to a page inside SPA. To fix that, we subscribe to the Angular router inside the app.component.ts file and send the new routes to Google Analytics.

PDR does not route to a page inside the app so it does not has this issue so far. But I added this piece of code here as a standard GA4 implementation.


RayPlante (approved):
Tested on testdata (as part of oar-docker 1.11.0rc2); no discernable issues on browser-side, and no errors appearing on the server-side.

PR #324, opened 2024-02-27
Handled missing pdrid in metrics

chuanlin2018:
Issue: https://sgitlab.nist.gov/oar-odd/landing-page-service/-/issues/3
Landing page crashed because there was no pdrid in metrics data.

Fix: Added pdrid check in getMetrics() in both landingpage and metrics component.

### DIFF
--- a/angular/src/app/landing/landingpage.component.ts
+++ b/angular/src/app/landing/landingpage.component.ts
@@ -300,36 +300,7 @@ export class LandingPageComponent implements OnInit, AfterViewInit {
             if(event.type == HttpEventType.Response){
                 this.recordLevelMetrics = JSON.parse(await event.body.text());
 
-                let hasFile = false;
-
-                if(this.md.components && this.md.components.length > 0){
-                    for(let com of this.md.components) {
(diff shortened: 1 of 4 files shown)


RayPlante (commented):
I tested this locally by ingesting a new record and then editing its record in recordMetrics to  remove the `pdrid` property.  I confirmed the error with the unpatched version of the app and then confirmed the fix with the patch from this PR.  

Just waiting on the remaining updates

chuanlin2018:
Just checked the fix for metrics page. Still working on landing page side... There are similar code in both page that need be put in a service but will work on it later.

chuanlin2018:
All fixed.

chuanlin2018:
In metrics page I moved file level metrics code inside record level metrics function because if there is no metrics level data, no need to fetch file level data.

chuanlin2018:
Loading metrics data after Nerdm record loaded.

RayPlante (approved):
Tested locally again; no issues found.

RayPlante:
Checked on testdata; no issues.  Merging.

PR #325, opened 2024-03-12
Hide download button for folders

chuanlin2018:
This check in fixed following issue:
Issue: In the file list, download button showed up for folders which should not.
The column title that sometime display "Status" and sometime "Access" was by design. The default name is "Status". But when access level is 'restricted public', it changes to "Access". Here is the code:
        if(this.record['accessLevel'] === 'restricted public') {
            this.checkAccessPageType();
            this.cols[3]['header'] = "Access";
        }
To test the app locally, set the following config in environment.ts:
    useMetadataService: true,
    useCustomizationService: true
And change data.nist.gov to testdata.nist.gov if necessary.

Test case: 
http://localhost:4200/lps/mds2-2909 (in testdata.nist.gov)
http://localhost:4200/lps/mds2-2645 (in data.nist.gov)

### DIFF
--- a/angular/src/app/landing/data-files/data-files.component.html
+++ b/angular/src/app/landing/data-files/data-files.component.html
@@ -26,8 +26,8 @@
                         <span><b id="filelist-heading">Files </b> </span>
 
                         <a class="faa-stack fa-lg icon-download" *ngIf="!editEnabled" (click)="downloadAllFiles()"
-                        data-toggle="tooltip" title="Download all files"
-                        [ngStyle]="{'color':getDownloadAllBtnColor(),'cursor':'pointer'}">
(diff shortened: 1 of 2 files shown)


RayPlante:
@chuanlin2018 (& @GRG2):  I see.  I don't remember what our reasoning was at the time, but it looks like we had some.  I think, though now, now that we see it, we prefer "Access" across the board.  Do you think you could make that change?

Meanwhile, I'm testing the download link removal for folders.

chuanlin2018:
Changed.

RayPlante (approved):
both updates work as expected (see on testdata)

chuanlin2018:
Confirmed in testdata.

PR #326, opened 2024-03-25
Modified landing page for semiconductors

chuanlin2018:
In this check in, filters component has been restructured. A Collection class was created to NIST, Forensics and Semiconductors collection. A new component - Taxonomy was created to display a given collection's taxonomy. Thus we can loop through the collection objects and display the taxonomy.

Currently only two taxonomy data are displayed for each collection. For example, Forensics only display standard NIST taxonomy and Forensics taxonomy. Semiconductors only display standard NIST taxonomy and semiconductors taxonomy. With the new structure, it's much easier to add more collections for display if necessary.

I am also trying to make the color scheme configurable for each collection. The filter component is working but the menu bar at the right side still hard coded.

To test it locally, use data.nist.gov as backend:
http://localhost:4200/lps/pdr0-0001     - for Forensics

Need sample data to test semiconductors. I modified Forensics data to be the fake semiconductor data for testing.


deoyani (approved):
It is all working.

RayPlante:
Please do not merge yet.

PR #327, opened 2024-04-03
Improve Error Handling and UI Feedback for Dataset Availability

elmiomar:
This PR improves error handling and UI feedback for dataset availability within the datacart. 

### Changes:

- Added a property to track success or failure: in the data cart component **isCartLoadedSuccessfully** is used to track if the cart was loaded successfully. **isCartLoadedSuccessfully** is set to true if the result is successfully received and it's not null or empty. If there's an error or the result is null, we set it to false.

- Added ***ngIf** directive to the **app-treetable** element: in the datacart component's HTML template, we use the ***ngIf** directive to conditionally render the **app-treetable** element based on the **isCartLoadedSuccessfully** property.

- Updated the error handling logic in the **getRpaCart()** of the datacart service:  when an error occurs, we check the error status and message. If it matches our criteria (a 404 status and the specific message about metadata not being found), we propagate a specific error message back to the component. In the datacart component, we handle this specific error by displaying a user-friendly message based on the error information received from the service.

- Used RxJS's **catchError** operator in the service to intercept 404 errors and convert them into observable streams with custom error messages.

### Testing

- Tested end-to-end using oar-docker locally; a warning message is displayed when the metadata is not found.

### DIFF
--- a/angular/src/app/datacart/cart.service.ts
+++ b/angular/src/app/datacart/cart.service.ts
@@ -76,6 +76,7 @@ export class CartService {
                 }, 
                 (err) => {
                     // Log the error or handle it if needed
+                    console.error("Error:", err);
                     console.error("Error in getRpaCart:", err.message);
                     subscriber.error(err.message); // Propagate the error to the component
                     subscriber.complete();
--- a/angular/src/app/datacart/cart.service.ts
+++ b/angular/src/app/datacart/cart.service.ts
@@ -76,17 +76,31 @@ export class CartService {
                 }, 
                 (err) => {
                     // Log the error or handle it if needed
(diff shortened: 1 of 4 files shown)


chuanlin2018:
If this branch is for non-rpa cases as well, http://localhost:4200/datacart/global_datacart does not appear to be working because the data tree component is hidden. Can you check?

chuanlin2018:
Suggest something like *ngIf="!isRpa || isCartLoadedSuccessfully" - add a boolean variable to tell if this is RPA record so it dosplay normally for non-rpa records.

chuanlin2018 (approved):
Tested locally. Appeared to be working.

PR #336, opened 2024-04-12
Fix preservation workflow bugs

RayPlante:
This PR attempts to fix some intermittent bugs that result in preservation failures that do not result in a failure notification.  Namely,
  * A preservation request would fail because the reference to the review directory would set to None
  * A preservation request would to launch asynchronously in a separate thread without raising a catchable exception.

As these bugs were difficult to debug in a dev environment, they were deployed as a hot fix to datapub.nist.gov.  Since then, preservation requests have been completing successfully, and neither of the above failures appear to have occurred.  (At the very least, the fix did not appear to break anything.)


RayPlante:
This was confirmed on via a hot fix on datapub.nist.gov.  Will merge after confirmation of CI unit tests.

PR #344, opened 2024-08-14
Fix/odd11 delimit topic tag

chuanlin2018:
https://sgitlab.nist.gov/oar-odd/landing-page-service/-/issues/11

This branch has much more changes than expected.

1. Use bubble display for topics and keywords.
2. Break topics into three categories: CHIPS Metrology, Forensics and NIST R&D.
3. Fixed edit mode due to the changes in #2.
4. Modified the taxonomy tree view
5. Fixed the vertical alignment problem of the dropdown boxes in the search result title bar.
6. Changed the style of the section titles

Tested locally with fake backend (but the real json file provided by Gretchen).

To test it locally, turn on fake backend service, npm run, and browse:
http://localhost:4200/lps/test3
http://localhost:4200/lps/pdr0-0002


chuanlin2018:
Fixed following issues:
1. Forensics banner showed up briefly before CHIPS banner for CHIPS Metrology page.
2. Display 20 item per page instead of 10 items.
3. Uncheck all filters when landing page starts
4. Incorrect search result when select Materials Research then select Metrology - handle special character "&" in taxonomy.

RayPlante (approved):
All changed were tested on oardev.

I did have to fix a build-related issue: the new code used a relatively new function of the standard `Array` class, `at()`.  To get code to build under docker, I needed to upgrade Node to 16.17.1 as well as a newer version of Chromium.  The latter was accomplished via a Debian upgrade to Bullseye.

PR #347, opened 2024-09-17
Disabled fake backend

chuanlin2018:
Two changes in this check in:
1. disable fake backend in app.module.ts
2. Added alert to fake backend interceptor.

Unable to test the bundle download function with fake backend disabled...


RayPlante (approved):
That fixed it!

PR #348, opened 2024-09-25
Fixed wrapping issue in forefox and Safari

chuanlin2018:
Fixed topic and keyword wrapping issue in Firefox and Safari.

Tested locally.


RayPlante (approved):
As best as I can remember, this fixes the wrapping issues we saw before.  

There is one strange, non-repeatable issue seen on the forensics collection page.  Sometimes when open a particular top-level Forensics topic in the left-hand filters (but not any topic and not all the time), the right-hand nav bar will disappear from view because it was sent all the way to the bottom of the page, overlapping the footer.  

Nevertheless, I'm going to approve these changes as being better than it was.

PR #349, opened 2024-09-27
Fixed search filter issue in result list panel

chuanlin2018:
The issue with the search box has been fixed.

Also the "Clear all" button will truly clear everything this time.

May want to deploy to oardev first to let more people play with it before go to production.


RayPlante (approved):
Confirmed that this works under oar-docker locally.  We'll deploy to testdata to ensure it is working correctly there.

RayPlante:
FYI, GA test failure was anonalous.  Confirmed passing under GA via test branch.

PR #351, opened 2024-11-14
Fixed Firefox issues

chuanlin2018:
This branch fixes two issues:
1. When expand/collapse filter items, white space added to the top area of landing page. This issue only happened in Firefox.
2. Search result return nothing if taxonomy tag contains comma.

Issue #1 was because of a Firefox bug related to sticky styling. The fix was to use absolute position and dynamic position.

Issue #2 was fixed by replacing filter string with object.

Tested locally in Chrome, Firefox and Safari. Not tested in docker because had some issues running docker locally even integration branch - rmm had problem accessing Mongodb.

### DIFF
--- a/angular/src/app/_helpers/fakeBackendInterceptor.ts
+++ b/angular/src/app/_helpers/fakeBackendInterceptor.ts
@@ -6,10 +6,17 @@ import { userInfo } from 'os';
 
 @Injectable()
 export class FakeBackendInterceptor implements HttpInterceptor {
-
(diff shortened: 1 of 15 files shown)


RayPlante (approved):
Tested on testdata, confirmed the follow fixed under firefox:
  * Grand Challenge filters work properly
  * Opening up a filter does not insert extra space at the top of the page
  * Research areas list wrap properly now.

PR #354, opened 2024-12-19
Rename Bundle Name

elmiomar:
This PR updates the logic for generating bundle names in the datacart to address Greg's feedback.

New name format: **NIST-Data-[timestamp]**, where timestamp is in a human readable format: `YYYY-MM-DDTHH-MM`.

### Changes:

1. Updated `generateZipFileName` Method

- The new implementation generates a file name based on the current timestamp:

```typescript
private generateZipFileName(base: string = "NIST-Data"): string {
    const now = new Date();
    const timestamp = now.toISOString().slice(0, 16).replace(":", "-");
    return `${base}-${timestamp}`;
}
```
2. Replaced Bundle Name Logic

- Updated references to use the new method:

```typescript
let bundleBaseName = this.generateZipFileName("NIST-Data");
```

### Testing:
- These changes were tested locally using `oar-docker` to verify that the bundle names align with the new format when downloading bundles from the datacart.

### DIFF
--- a/angular/src/app/datacart/bundleplan/bundleplan.component.ts
+++ b/angular/src/app/datacart/bundleplan/bundleplan.component.ts
@@ -307,7 +307,7 @@ export class BundleplanComponent implements OnInit {
 
         // Sending data to _bundle_plan and get back the plan
         this.downloadFiles = this.dataCart.getSelectedFiles();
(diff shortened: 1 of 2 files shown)


chuanlin2018 (approved):
Tested locally and worked as expected.

PR #355, opened 2025-01-21
Set data cart zip file names to be more human-oriented

RayPlante:
This PR is part of an effort to make zip files created via the data cart more human-oriented; this one addresses the name assigned to zip files containing RPA data.


RayPlante (commented):
tested on testdata.

PR #358, opened 2025-04-10
Fix URLs linking to bulkdownload page to work under oar-docker

RayPlante:
For various reasons, the nginx URL mapping is not able to properly map the links to the bulkdownload page; thus, the links would not work under oar-docker.  To address this in this PR, absolute URLs are formed using base URLs pulled from the configuration.  (Similarly, the URL for downloading the script is now formed in a similar way.)

Also fixed is the rendering of the script execution examples on the bulkdownload page which were not showing up when accessed via the `/bulkdownloads` route (i.e. without an id included).  

Finally, two GitHub Action config files were updated to upgrade from a deprecating version of the cache action.

### DIFF
--- a/.github/workflows/angular-source.yml
+++ b/.github/workflows/angular-source.yml
@@ -22,7 +22,7 @@ jobs:
           node-version: 16.17
 
       - name: Setup dependency caching
-        uses: actions/cache@v2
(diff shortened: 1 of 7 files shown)


chuanlin2018 (approved):
Looks good to me.

PR #361, opened 2025-04-18
Load X icon from local; Enlarge NIST logo in footer.

chuanlin2018:
1. Load X icon from local.
2. Enlarge NIST logo in footer.
3. Some minor layout adjustment.


RayPlante (approved):
Needed to change URL to x--logo.svg to be relative to work with oar-docker.  Works as needed

PR #362, opened 2025-04-21
Fix/odd15 metrics02

chuanlin2018:
I am sure this will read data from metadatatransfer. When testing locally, it did trigger the metadataservice.

I am trying to run it in docker. Since I haven't run it for a while, might take a while to get it up and running properly.

### DIFF
--- a/angular/src/app/metrics/metrics.component.ts
+++ b/angular/src/app/metrics/metrics.component.ts
@@ -128,18 +128,20 @@ export class MetricsComponent implements OnInit {
                     this.record = md as NerdmRes;
                     this.datasetTitle = md['title'];
                     this.pdrid = md['@id'];
-    
-                    this.createNewDataHierarchy();
-                    if (this.files.length != 0){
-                        this.files = <TreeNode[]>this.files[0].data;
-                    }else{
-                        this.noChartData = true;
-                    }
-    
-                    this.expandToLevel(this.files, true, 0, 1);
 
-                    //Fetch metrics data regardless if there is chat data or not
-                    this.getMetricsData();
+                    if (this.inBrowser) {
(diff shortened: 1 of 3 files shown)


angular/src/app/metrics/metrics.component.ts
```diff
@@ -113,17 +118,18 @@ export class MetricsComponent implements OnInit {
...
+            // Get dataset
+            this.mdserv.getMetadata(this.ediid).subscribe(
+                (md) => {
+                    // successful metadata request
+                    this.record = md as NerdmRes;
+                    this.datasetTitle = md['title'];
+                    this.pdrid = md['@id'];
+
+                    if (this.inBrowser) {
                         this.createNewDataHierarchy();
```
RayPlante, line 133: I added this `inBrowser` guard to skip retrieving the metrics data when on the server-side.

RayPlante (approved):
This was throwing errors on the server-side because the configured URL for fetching the metrics data doesn't work on the server-side.  (The failure was causing the "Oops" message to appear briefly on the client-side.)  I put in a fix for this (see below).  

There are two ways to fix this issue.  One is by adding the proper server-side URL for getting the metrics data into configuration (under `APIs.server_side`).  However, when the number of files is large, this would add significant delay while the server-side renders all the information.  Instead, I put in a `inBrowser` guard to skip the fetching on the server-side.  

@chuanlin2018, please confirm that this does not have any ill side effects.

angular/src/app/metrics/metrics.component.ts
chuanlin2018: Oh yes. I overlook this block. Just tested locally and it works.

